PR #4, opened 2017-11-17
Adding static validation for SqlPS installation

niadak:


images/Windows/Vs2017/scripts/Install-SQLPowerShellTools.ps1
```diff
@@ -52,4 +52,9 @@ $SharedManagementObjectsExitCode = InstallMSI -MsiUrl "https://download.microsof
 
 $PowerShellToolsExitCode = InstallMSI -MsiUrl "https://download.microsoft.com/download/8/7/2/872BCECA-C849-4B40-8EBE-21D48CDF1456/ENU/x64/PowerShellTools.msi" -MsiName "PowerShellTools.msi"
 
+$env:PSModulePath = $env:PSModulePath + ";C:\Program Files\Microsoft SQL Server"
```
chrisrpatterson: I would much rather see us read the PSModulePath back from the machine level key for validation.  For example https://github.com/Microsoft/vsts-image-generation/blob/9722590ca1c609b1bf19d5225bcbf67d55a7d62b/images/Windows/Vs2017/scripts/Install-Python.ps1#L19 for how we are validating python is in the machine path.

PR #7, opened 2017-11-27
Refactored powershell scripts

lokesh755:


images/win/Vs2017/scripts/Install-NodeLts.ps1
```diff
@@ -19,9 +19,6 @@ Add-MachinePathItem $PrefixPath
 Add-MachinePathItem $CachePath
```
chrisrpatterson: I don't think this needs to be in the path only the prefix does.  We should also set NPM_CONFIG_CACHE and NPM_CONFIG_PREFIX as system level environment variables with the appropriate values.

images/win/Vs2017/scripts/Validate-ServiceFabricSDK.ps1
```diff
@@ -0,0 +1,18 @@
+################################################################################
```
chrisrpatterson: @NCarlsonMSFT can you review this.

images/win/Vs2017/scripts/Install-ServiceFabricSDK.ps1
```diff
@@ -12,5 +12,3 @@ New-Item -Path $temp_install_dir -ItemType Directory -Force
 Set-ExecutionPolicy RemoteSigned -Scope CurrentUser -Force
 
 WebpiCmd.exe /Install /Products:MicrosoftAzure-ServiceFabric-CoreSDK /AcceptEula
```
chrisrpatterson: @NCarlsonMSFT  can you review this

NCarlsonMSFT: Double checked that is the right product ID

PR #32, opened 2018-01-08
adding batch file to uninstall existing azure powershell modules inst…

arjgupta:
…alled through installer


lokesh755:
Not merging the current pull request , because the script is calling Install-Azuremodules.ps1. But I have added the uninstall command to the template in different commit. https://github.com/Microsoft/vsts-image-generation/commit/055502dfbdbca17c4e0b4997b333d6e2884ecc38

PR #79, opened 2018-02-22
Go versions 1.9.4 and 1.10 with env vars pointing to locations

DavidStaheli:
Go version 1.9.x and 1.10.x are the 2 most recent minor versions of Go.  The plan is to have the 2 latest minor releases on Hosted VS2017 and Hosted Linux.


lokesh755:
From the next releases we are trying to generate Readme.md while running packer process. Can you update Validate-Go.ps1 to reflect the changes you've added to Readm.md where I've added code to generate markdown ? 

https://github.com/Microsoft/vsts-image-generation/commit/6ead759e98568afc7feacadfa6626a4f7149ba0a#diff-ba5762ecba240c4b1f7b278cdb7c4910

DavidStaheli:
Hi @lokesh755, that is very cool.  Thanks for doing it for the validators I've been working on.  In addition to listing the tool version that is in the path (from running `go version`), is there a recommended way to add to the Markdown the tools that are not in the path?  I'd like the Markdown to note the versions of Go, Python, Ruby, Java, etc. that are installed, but not the default version in the path.

lokesh755:
I'm updating $env:Path to check the versions of software which are not in path by default, since it's valid for only that powershell session. You can see the similar example on python validation script. https://github.com/Microsoft/vsts-image-generation/blob/master/images/win/scripts/Installers/Validate-Python.ps1#L21

DavidStaheli:
@lokesh755 thanks for the tip.  This PR is ready for review again.

images/win/scripts/Installers/Install-Go.ps1
```diff
@@ -1,28 +1,63 @@
...
+    Remove-Item go$goVersion.windows-a`md64.zip
+
+    # Make this the default version of Go?
+    if ($addToDefaultPath)
+    {
+        Write-Host "Adding Go $goVersion to the path..."
+        # Add the Go binaries to the path.
+        Add-MachinePathItem "C:\$newDirName\bin"
+        # Set the GOROOT environment variable.
+        setx GOROOT "C:\$newDirName" /M
```
lokesh755: Can you change the above statements to 
` Add-MachinePathItem "C:\$newDirName\bin" | Out-Null`
`$output =  setx GOROOT "C:\$newDirName" /M | Out-Null`

Because when the current function is called with with -addToDefaultPath  as $True, output from the above statements are in the output pipeline which might not be intended.

PR #82, opened 2018-02-24
Ruby 2.3.3, 2.4.3-1, 2.5.0-1 installation and validation

DavidStaheli:


images/win/scripts/Installers/Install-Ruby.ps1
```diff
@@ -0,0 +1,65 @@
...
+    $installDirectory = "C:\Ruby$rubyVersion-$rubyArch"
+    $installerName = "rubyinstaller-$rubyVersion-$rubyArch.exe"
+    $argumentList = ("/dir=$installDirectory", "/verysilent", "/tasks=nomodpath,noassocfiles,noridkinstall,defaultutf8")
+    Install-EXE -Url $installerUri -Name $installerName -ArgumentList $argumentList | Out-Null
+
+    # Add to default path?
+    if ($addToDefaultPath)
+    {
+        Write-Host "Adding $installDirectory\bin to the default path..."
+        Add-MachinePathItem "$installDirectory\bin"
```
lokesh755: Can you change the above statement to
`Add-MachinePathItem "$installDirectory\bin" | Out-Null`

Because when the current function is called with with -addToDefaultPath as $True, output from the above statement is in the output pipeline which might not be intended.

DavidStaheli:
Thanks for the helpful catch, @lokesh755.

PR #90, opened 2018-03-07
Porting web browser changes to vs2015 image

kaadhina:


kaadhina:
@lokesh755 @cltshivash

Issue #91, opened 2018-03-07
Install-JavaTools.ps1 is failing


lokesh755:
Latest install of maven 3.5.2.2 removed "M2" and "Maven_opts" keys from the registry which are being used to set the environment variables. Setting environment variables is failing due to unavailability of keys.

PR #92, opened 2018-03-07
Update Install-JavaTools.ps1 with maven variables

kelliejos:
Update to installer to set variables removed from current maven chocolatey script.


lokesh755:
Resolves  #91

PR #97, opened 2018-03-14
Update for the current LTS version installed by Chocolately

DavidStaheli:
**Question:** Do version updates in this file need to wait until the versions actually hit production?

### DIFF
--- a/images/win/Vs2017-Server2016-Readme.md
+++ b/images/win/Vs2017-Server2016-Readme.md
@@ -358,7 +358,7 @@ _Environment:_
 
 ## Node.js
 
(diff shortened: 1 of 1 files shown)


lokesh755:
Usually we wait till the image hit all production except this time. Btw the current image has v8.9.4 not v8.9.10. We generated the image on March 7 the day before v8.9.10 got approved on Chocolatey.

DavidStaheli:
Thanks, @lokesh755.  The next time the image is generated, it will have 8.10.0 as this PR attempted to show, just FYI.

PR #109, opened 2018-04-02
Disable Git autocrlf

brcrista:
The Git autocrlf setting is getting set automatically by the Chocolatey installer for Git.  This means that all files in the repo that get copied to artifacts will have their line endings tampered with.  This is significant when the repo contains things like Linux shell scripts that should not have their line endings changed.


brcrista:
On second thought, I'm not sure we want to do this.  This could conversely break users who are developing on Windows and may be relying on the build machine to have autocrlf set.

brcrista:
Users can use .gitattributes to make sure their line endings are handled in a particular way on CI builds.

MSP-Greg:
Obviously, closed a while ago, but:

> This could conversely break users who are developing on Windows and may be relying on the build machine to have autocrlf set.

As stated in the above mentioned issue, one would normally assume the build machine is not changing anything.  This caused issues with a repo that tested fine on Appveyor.

brcrista:
@MSP-Greg I am talking about back compat.  If we were starting clean I think no autocrlf would be the right choice.  But now that we have shipped this way, we can't change it without potentially breaking people.

MSP-Greg:
Understood.

pombredanne:
@MSP-Greg you wrote:

> As stated in the above mentioned issue, one would normally assume the build machine is not changing anything. This caused issues with a repo that tested fine on Appveyor.

Same here. This is rather weird when tools are not using their defaults.  Weird is an understatement.

PR #111, opened 2018-04-03
Removed default Ruby; Added Ruby to tools cache

lkillgore:


images/win/scripts/Installers/Download-ToolCache.ps1
```diff
@@ -18,4 +18,16 @@ AzCopy /Source:$SourceUrl /Dest:$Dest  /S /V /Pattern:$Path
...
 
+$current = Get-Location
+Set-Location -Path $ToolsDirectory
+
+Get-ChildItem -Recurse -Depth 4 -Filter install_to_tools_cache.bat | ForEach-Object {
+    Write-Host $_.DirectoryName
+    Set-Location -Path $_.DirectoryName
+    Get-Location | Write-Host
+    Expand-Archive 'tool.zip' -DestinationPath '.'
+    cmd.exe /c 'install_to_tools_cache.bat'
```
brcrista: Proposal: can we just name this `setup.cmd` to parallel the Linux / macOS tool cache?  (also this: https://stackoverflow.com/questions/5347593/whats-the-difference-between-bat-and-cmd-file)

lkillgore: I'm concerned about name collisions.  There could easily be a tool with a setup.cmd or setup.bat.  I did (afterward) add a depth of 4, so likely not an issue anymore--for Windows, but it may be for Linux.  What are the possibilities that a tool will have a `setup.sh` file?  I'm on board with renaming it to be consistent; I'm not sure setup.cmd is the best idea.

images/win/InstalledSoftware.md
```diff
@@ -0,0 +1,445 @@
+# VSTS Hosted VS2017 image
```
lokesh755: Remove InstalledSoftware.md. It's generated when you run the packer process for reference

images/win/Vs2017-Server2016-Readme.md
```diff
@@ -296,27 +296,6 @@ _Environment:_
 * GOROOT: root directory of the Go 1.10 installation
 * GOROOT_1_10_X64: root directory of the Go 1.10 installation
 
-## Ruby (x64)
```
lokesh755: Do not edit the readme.md file. We generally update it with generated readme whenever we push corresponding image to Hosted VS2017 pool

images/macos/macos-Readme.md
```diff
@@ -17,7 +17,6 @@ The following software is installed on machines in the VSTS **Hosted macOS Previ
 - PowerShell 6.0.1
 - Python 2.7.10
 - Python 3.6.4
-- Ruby 2.5.0p0
```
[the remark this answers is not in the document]
lkillgore: Good catch.  I was under the impression that this file was not used, but upon closer inspection it is being edited.  I've restored it.

images/win/scripts/Installers/Download-ToolCache.ps1
brcrista: I wasn't too concerned about that.  If a tool has its own file named setup.sh / setup.cmd file and we want it to get run, then it gets run.  If we don't want it run, then we shouldn't be distributing it in the tools cache in the first place (unless there's some weird situation where a tool has a "setup.sh" that's not actually used for first-time setup).

If we want to be safer about this, we can put the setup file on the same level as the x64 directory and x64.complete file in the tools cache.  That might make more sense so we don't pollute the tools directory.

lkillgore: I did a quick google search, and both file names seem to be reasonably popular (increasing the likelihood for name collisions in the future).  Putting them beside the x64 folder would be fine if we don't also search the sub-folders.

PR #112, opened 2018-04-05
Update Install-DACFx.ps1

Ajay-MS:
Updated DacFx to latest version


Ajay-MS:
SQL team has fixed few issues in the latest version of the data-tier application framework and hence upgrading it to the latest version.

PR #113, opened 2018-04-06
Install MinGW

brcrista:
MinGW contains GNU build tools -- most importantly GCC and Make.  Both of these are needed for C and C++ support as well as Python support.

### DIFF
--- a/images/win/scripts/Installers/Install-MinGW.ps1
+++ b/images/win/scripts/Installers/Install-MinGW.ps1
@@ -4,7 +4,7 @@
 ##  Desc:  Install GNU tools for Windows to C:\tools\mingw64
 ################################################################################
 
-Import-Module -Name PathHelpers -Force
+Import-Module -Name ImageHelpers -Force
 
 choco install -y mingw
 $installDir = "C:\tools\mingw64\bin"
--- a/images/win/scripts/Installers/Install-MinGW.ps1
+++ b/images/win/scripts/Installers/Install-MinGW.ps1
@@ -1,0 +1,15 @@
+################################################################################
+##  File:  Install-MinGW.ps1
+##  Team:  CI-X
+##  Desc:  Install GNU tools for Windows to C:\tools\mingw64
+################################################################################
+
+Import-Module -Name PathHelpers -Force
+
+choco install -y mingw
+$installDir = "C:\tools\mingw64\bin"
+Add-MachinePathItem $installDir
+
+# Hard link mingw32-make.exe to make.exe, which is a more discoverable name
+# and so the same command line can be used on Windows as on macOS and Linux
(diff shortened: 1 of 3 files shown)


images/win/scripts/Installers/Install-MinGW.ps1
```diff
@@ -0,0 +1,15 @@
...
+
+Import-Module -Name PathHelpers -Force
+
+choco install -y mingw
+$installDir = "C:\tools\mingw64\bin"
+Add-MachinePathItem $installDir
+
+# Hard link mingw32-make.exe to make.exe, which is a more discoverable name
+# and so the same command line can be used on Windows as on macOS and Linux
+New-Item -Path $installDir -ItemType HardLink -Name "make.exe" -Value "$installDir\mingw32-make.exe"
```
AlexRukhlin: What about gcc? Shouldn't we add a hard link for it as well?

images/win/scripts/Installers/Validate-MinGW.ps1
```diff
@@ -0,0 +1,41 @@
+if (Get-Command -Name 'g++')
```
lkillgore: Better to use `gcc` or `g++`?

brcrista: I don't think it matters much.  There are a bunch of tools installed as part of the MinGW suite, so I just chose one somewhat arbitrarily to check.

image

bryanmacfarlane: Even though it's sort of arbitrary, we should ask ourselves why did we install this on the image in the first place??  I assume it's gcc.   If that's the case, validating it's in the PATH and it runs (print gcc version) would be a good thing.  That's what we're after here - static validation and catch install issues early.  E2E validation is a job for buildcanary.

brcrista: The main use cases are for make, and gcc, and g++.  I can add a check for gcc.

images/win/scripts/Installers/Install-MinGW.ps1
```diff
@@ -0,0 +1,15 @@
+################################################################################
+##  File:  Install-MinGW.ps1
+##  Team:  CI-X
+##  Desc:  Install GNU tools for Windows to C:\tools\mingw64
+################################################################################
+
+Import-Module -Name PathHelpers -Force
```
lokesh755: Name of the module is not "PathHelpers" , it's "ImageHelpers"

brcrista: I was importing the PathHelpers file directly, but I can import the whole module instead.

PR #117, opened 2018-04-16
Validate the tools cache as part of the build

brcrista:
We need to have a build step to make sure the hosted tools cache contains the tools we expect.  It will be the responsibility of each tool owner to add build-time checking for their tool.


images/win/scripts/Installers/Validate-ToolCache.ps1
```diff
@@ -0,0 +1,37 @@
...
+    }
+    else
+    {
+        Write-Host "Python $v ($arch) is not installed"
+        exit 1
+    }
+
+    if (Test-Path "$path\Scripts\pip.exe")
+    {
+        Write-Host "Pip for Python $v ($arch) is successfully installed:"
```
lkillgore: I'm not a fan of the language `is successfully installed`; I'd prefer something like `detected Python X.X`.  YMMV.

brcrista: Well, this is to match what all the other Validate-* scripts say.  Personally I would say "installed" though :)

brcrista:
@lokesh755 could you review when you have a chance?  Looks like you'll have to be the one to complete the merge.

PR #120, opened 2018-04-19
Add x86 Python versions to validation

brcrista:
I am adding x86 Python versions to the hosted tool cache on Windows.


brcrista:
@lokesh755

PR #126, opened 2018-04-27
Fix for intermittent sysprep issues

NateB2:
When building the Windows Server 2016 VS2017 image, I've been running into intermittent sysprep issues where the image appears to have been created successfully, but when the image is used for a VM, the VM does not start successfully.  This has happened both with VHD images and with managed images (I'm using the VM size "Standard_F8s_v2" to build the images).  Changing the sysprep step to adhere to Microsoft's latest recommendations for creating a managed image appears to fix the issue.  Instead of shutting down immediately after running sysprep, it loops until the image state is correct.  Here is the example where I copied this change from: https://github.com/hashicorp/packer/blob/master/examples/azure/windows_custom_image.json


chrisrpatterson:
Looks good to me @lokesh755 can you take a look and merge.

PR #156, opened 2018-06-04
Users/logopu/windowscontainer1803

lokesh755:


veyalla:
hi @bryanmacfarlane and @lokesh755 - any idea when this will be merged. We need this for generating Azure IoT Edge docker images for RS4/1803. Thanks!

bryanmacfarlane:
@lokesh755 - let's discuss tomorrow.  As long as we have canary coverage for the minimal set of tools on the host, we should be good.

PR #175, opened 2018-07-12
Corrected .NET472 download location.

ignatz42:
Validate-NET472.ps1 fails with "vhd: .Net 472 not found" after Install-NET472.ps1 fails with "vhd: Failed to install the Executable NDP472-DevPack-ENU.exe
vhd: The remote server returned an error: (404) Not Found." This change updates the location where the installer lives.


ryanspletzer:
I'm running into this issue as well. Can someone approve this PR?

PR #177, opened 2018-07-14
Enable TLS 1.2 before downloading from dev.mysql.com

ignatz42:
Install-MysqlCli.ps1 fails with the error message:
"vhd: Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel."


duk3ofed:
@ignatz42 I too have run into this problem. Given that more public endpoints will likely require TLS v1.2 before long, however, wouldn't it be better to fix this somewhere further up the chain, rather than only in the mysqlcli script? For example, in the "initialise-VM" script perhaps? Since this TLS setting seems to be per shell session as well, it might be an idea to add to the $profile.currentuserallhosts profile as part of the process - that way when each individual installer script is initialised, it will enable TLS v1.2 for all of them. Given that many endpoints still rely on <TLS1.2, the code should be "[Net.ServicePointManager]::SecurityProtocol = 'tls12, tls11, tls'"

ryanspletzer:
I too am facing this issue.

lokesh755:
Thanks for the PR!

PR #201, opened 2018-08-21
confirmation password >14 and test install_user part of admin group

riezebosch:
To prevent from this error:
vhd: The password entered is longer than 14 characters.  Computers
vhd: with Windows prior to Windows 2000 will not be able to use
vhd: this account. Do you want to continue this operation? (Y/N) [Y]:

Because it fails the packer build later on in the process when scripts are installed as an elevated user.


lokesh755:
Thanks for contributing!

PR #209, opened 2018-08-28
Adding Android SDK to Ubuntu 16.04

joshmgross:
Adding the Android SDK and build tools to Ubuntu 16.04 for parity with Windows VS2017 agent


images/linux/scripts/installers/android.sh
```diff
@@ -0,0 +1,67 @@
...
+    "add-ons;addon-google_apis-google-24" \
+    "add-ons;addon-google_apis-google-23" \
+    "add-ons;addon-google_apis-google-22" \
+    "add-ons;addon-google_apis-google-21" \
+    "cmake;3.6.4111459" \
+    "patcher;v4"
+
+# Document what was added to the image
+echo "Lastly, documenting what we added to the metadata file"
+DocumentInstalledItem "Android SDK [${ANDROID_ROOT}]"
```
chrisrpatterson: We should document all of the installed versions like https://github.com/Microsoft/vsts-image-generation/blob/master/images/win/scripts/Installers/Update-AndroidSDK.ps1#L124

images/linux/scripts/installers/android.sh
```diff
@@ -0,0 +1,67 @@
...
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+source $HELPER_SCRIPTS/apt.sh
+
+# Set env variable for SDK Root (https://developer.android.com/studio/command-line/variables)
+ANDROID_ROOT=/usr/local/bin/android
+ANDROID_ROOT_SDK=${ANDROID_ROOT}/sdk
+echo "ANDROID_ROOT_SDK=${ANDROID_ROOT_SDK}" | tee -a /etc/environment
```
chrisrpatterson: Should this be ANDROID_HOME

https://github.com/Microsoft/vsts-image-generation/blob/master/images/win/scripts/Installers/Update-AndroidSDK.ps1#L124

joshmgross: ANDROID_HOME is deprecated, see: https://developer.android.com/studio/command-line/variables

images/linux/scripts/installers/android.sh
joshmgross: I was worried about cluttering up https://github.com/Microsoft/vsts-image-generation/blob/master/images/linux/Ubuntu1604-README.md by adding every specific version

chrisrpatterson: People wan to know that level of detail.  It is important.

dakale: We typically document every version even if theres multiples, see go, java, php etc

images/linux/scripts/installers/android.sh
chrisrpatterson: ok then we should fix the other images @lokesh755

lokesh755: Sure!

PR #210, opened 2018-08-29
Update minor versions of Go for new releases

DavidStaheli:
Go 1.10.4 and 1.11 were just released.  This also sets Go 1.11 as the default Go version since it is the latest.


images/linux/scripts/installers/go.sh
```diff
@@ -31,5 +31,5 @@ function InstallGo () {
 
 # Install Go versions
 InstallGo 1.9 1.9.7 false
-InstallGo 1.10 1.10.3 true
-InstallGo 1.11 1.11rc2 false
+InstallGo 1.10 1.10.4 true
+InstallGo 1.11 1.11 false
```
bryanmacfarlane: This is good.  Note that the previous addition of 11rc2 goes against our hosted image policy of released versions

DavidStaheli: Thanks @bryanmacfarlane, I didn't know about that.

PR #221, opened 2018-09-05
Add leiningen to Ubuntu VM image

SergeySmolin:
This PR adds Leiningen to Ubuntu VM image

The Leiningen installation was tested and validated using the following steps.
1. Image generation script was updated to add Leiningen to VM.
2. VM image with Leiningen was successfully generated.
3. VM created from VM image with Leiningen
4. Leiningen works correctly on VM (*lein* command executes as expected)


max-lobanov:
@dakale , @DavidStaheli , Could you please look at it?

bryanmacfarlane:
@max-lobanov - did you create an issue?  See

https://github.com/Microsoft/vsts-image-generation#issues-and-requests

SergeySmolin:
Here is the issue for this PR: https://github.com/Microsoft/azure-pipelines-image-generation/issues/231

SergeySmolin:
There seem to be difficulties in getting leiningen not download on the VM, because leiningen downloads the jar to the directory specified by $LEIN_HOME environment variable, or if that is not set, by $HOME/.lein. $LEIN_HOME is not set during the VM runtime so it defaults to $HOME/.lein which does not have the jar even though we downloaded it during the provisioning. We are trying to resolve this, meanwhile any suggestions how to resolve this are greatly appreciated.

dakale:
When you say:
> LEIN_HOME is not set during the VM runtime so it defaults to $HOME/.lein which does not have the jar even though we downloaded it during the provisioning

Do you mean you are setting `$LEIN_HOME` during the install script, but it is not set when you run a build on the resulting image, causing `lein` to be redownloaded? 

You should be able to set `$LEIN_HOME` to eg `/usr/local/lib/lein` to have it installed there during provisioning. And then see `java-tools.sh` for an example, but also add `$LEIN_HOME` to `/etc/environment`. That will get picked up when it is run in a hosted context, but for testing you can manually set `$LEIN_HOME` to the same value before you call it in a script, or 

```
echo "##vso[task.setvariable variable=LEIN_HOME;/usr/local/lib/lein"
```
for eg if you want it to persist across tasks

SergeySmolin:
Hi @dakale, I have tried your suggestion about putting `$LEIN_HOME` into /etc/environment. When I then created a VM in a test resource group in Azure, `$LEIN_HOME` was not available in the environments. I think you meant when you said:
> That will get picked up when it is run in a hosted context

That for the test set up I described above it will not work?
But I can still do like this because it will work for the main use case? (hosted context - that is when the VM runs user builds in production?).

Also can you please clarify where should I use the following command:

```
echo "##vso[task.setvariable variable=LEIN_HOME;/usr/local/lib/lein"
```
I have difficulties understanding. Thank you.

SergeySmolin:
There is also another way. It seems like `/` directory on the VM is used as home directory so doing like this during provisioning
```
mkdir /.lein
curl -s https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein > /usr/local/bin/lein \
  && chmod 0755 /usr/local/bin/lein
LEIN_HOME=/.lein lein
```
allows lein to work right away without a download.
Is this an acceptable approach?

SergeySmolin:
Hi, @dakale. We updated PR based on your comments. We removed `Leiningen` from the `Ubuntu1604-README.md`. We added logic to install the Leiningen main script to the system wide directory. We added a `command -v` for validation.  We overrode `LEIN_HOME` env variable the same way as in `java-tools.sh`. We tested it and it looks good.
Please let us know if you have any comments/concerns.

dakale:
LGTM. Ill merge and keep the issue tracking the deployment to hosted pools open until an image with this installed makes it out. 

Thanks

PR #233, opened 2018-09-12
Add Python and Ruby from hosted tool cache to README

brcrista:


dakale:
We dont typically update the README direectly. I went ahead and added a better starting point for getting this information documented.

brcrista:
Thanks!

PR #244, opened 2018-09-18
Update installed versions of Go

DavidStaheli:
It's expected that we can remove Go 1.9 after Go 1.11 has been out for a few more weeks (it's brand new).


DavidStaheli:
Done!  Sorry about that.

PR #248, opened 2018-09-20
Warmup dotnet sdks

chrisrpatterson:
Change the process to install by SDK version in order and then warmup each sdk.

Will need to figure out how to preserve the sentinel files that are persisted in the userhome under .dotnet or .nuget/packages for this to really work.


chrisrpatterson:
@lokesh755 take a look and see if you think we  can make this work.  I think this will make dotnet faster.

PR #255, opened 2018-09-25
Update Vs2017-Server2016-Readme.md

jongio:
- Adding Windows Containers Only


DavidStaheli:
Since the readme files are now automatically generated by installation+validation scripts, the text that this PR adds to the readme shouldn't be added directly, but instead should be output at the bottom of this script:  https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Validate-Docker.ps1

PR #264, opened 2018-10-03
Document credentials for MySQL

vtbassmatt:
fixes https://github.com/MicrosoftDocs/vsts-docs/issues/1673


dakale:
@lokesh755 Is OOF, need to confirm with him, but I do not see in the Windows installer where the MySQL server is actually installed/configured, so Im not sure if it is even set up with a password. 

For Ubuntu, root/root is correct, and documented so we should make sure we have consistency ofc

Issue #276, opened 2018-10-10
(VS2017) Add jq command line json processor


DavidStaheli:
(VS2017) Add jq command line json processor

Issue #273, opened 2018-10-10
(VS2017) Add Mercurial


DavidStaheli:
(VS2017) Add Mercurial

Issue #260, opened 2018-09-29
7-Zip please consider adding to all images


MSP-Greg:
Please consider installing 7-Zip on all images.

I know there's a task that makes it available, but I wanted to use it in a ps1 file.  Maybe I need to go AFK, but I couldn't get the 7-Zip exe or msi files to install.  Might have something to do with the shell integration, I don't know.

I often use ps1 files for more involved CI tasks, and they also make it easier for people to run builds/tests locally.  Hence the task DSL would just clutter things up.

Besides, I suspect that an awful lot of developers use it, and having it unavailable for CI is a PITA.  I use it all the time with the other CI to bundle artifacts, etc.  They've even got an environment variable for it's full path...

PR #302, opened 2018-10-15
Install 7zip, jq, and Mercurial

brcrista:
#260 
#273 
#276


ihnorton:
Did the VS2015 image get this update? It seems to be missing `7z`:

```
7z : The term '7z' is not recognized as the name of a cmdlet, function, script 
file, or operable program. Check the spelling of the name, or if a path was 
included, verify that the path is correct and try again.
```

brcrista:
No, we're not adding software to that image anymore.  Are you using it for VS2015's C++ compiler?

ihnorton:
Good point, I will switch to the VS2017 image and set the toolchain instead.

PR #305, opened 2018-10-17
incorrect symlink

sharma-abhay:
Script creates a symlink on line 35 (ln -s /usr/share/apache-maven-3.5.4/bin/mvn /usr/bin/mvn).

This is incorrect and is causing all our maven builds on Hosted Ubuntu 1604 in Azure DevOps to fail. The error in Azure DevOps is

UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Unable to locate executable file: '/usr/share/maven/bin/mvn'. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.

I've found the root cause: Script sets M2_HOME to /usr/share/maven but symlink for maven executable is create at /usr/bin/mvn.


dakale:
Thanks. I fixed this in a different commit to preserve M2_HOME pointing to the location of the install instead

PR #310, opened 2018-10-18
Added ANDROID_HOME Env variable to Ubuntu

joshmgross:
Adding the ANDROID_HOME Environment variable since some older versions of Gradle and Maven rely on it instead of using the newer ANDROID_SDK_ROOT. 

See https://developer.android.com/studio/command-line/variables for more info.


madhurig:
@janpio - This change is included in this month's update. We are preparing for rollout next week and we expect it be available for all users by end of November.

janpio:
Thanks @madhurig. Good change, so much documentation online still uses `ANDROID_HOME`.

Issue #281, opened 2018-10-10
(VS2017) Add Perl


DavidStaheli:
(VS2017) Add Perl

DavidStaheli:
Completed by #311.

Issue #279, opened 2018-10-10
Add Nullsoft scriptable install system (NSIS)


DavidStaheli:
Add Nullsoft scriptable install system (NSIS)

riezebosch:
`Validate-NSIS.ps1` keeps failing and I don't know why:
```
2018-11-08T01:06:49.9150941Z     vhd: + if (Get-Command -Name makensis)
2018-11-08T01:06:49.9160465Z     vhd: +     ~~~~~~~~~~~~~~~~~~~~~~~~~~
2018-11-08T01:06:49.9179032Z     vhd:     + CategoryInfo          : ObjectNotFound: (makensis:String) [Get-Command], CommandNotFoundException
2018-11-08T01:06:49.9195986Z     vhd:     + FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand
```

When I execute the command locally everything seems to be okay.

x2bool:
'makensis' is not recognized as an internal or external command.

Seems like makensis is not in the PATH for some reason

riezebosch:
Just master indeed.

riezebosch:
I probably found the problem. This script:
```
echo $env:Path
gci "${env:ProgramFiles(x86)}\NSIS\"

$SoftwareName = 'Nullsoft Install System (NSIS)'

if (Get-Command -Name makensis)
{
    Write-Host "$SoftwareName is installed"
}
else
{
    Write-Host "$SoftwareName is not installed"
    exit 1
}
```

produces the following output:
```
018-11-20T08:43:50.0788475Z ##[command]"C:\Program Files\PowerShell\6\pwsh.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'C:\agent1\_work\_temp\2394b3c5-df60-49b2-8882-be237a75166e.ps1'"
[...]

Issue #278, opened 2018-10-10
(VS2017) Add OpenSSL 1.0.2L (64-bit)


DavidStaheli:
(VS2017) Add OpenSSL 1.0.2L (64-bit)

Issue #277, opened 2018-10-10
Add GitVersion 4.0.0 (or latest) (Windows)


DavidStaheli:
(VS2017) Add GitVersion

DavidStaheli:
This was completed by PR #311.

PR #311, opened 2018-10-19
Adding Perl, OpenSSL, GitVersion, NSIS, and CF CLI to VS2017 Image

joshmgross:
See #281, #278, #277, #279 for their respective issues


images/win/scripts/Installers/Validate-GitVersion.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+# Adding description of the software to Markdown
+$SoftwareName = "GitVersion"
+$version = $command.Version.ToString()
+
+$Description = @"
+_Version:_ $version<br/>
+"@
+
+#Write-Host $Description
```
brcrista: nit: remove

images/win/scripts/Installers/Validate-NSIS.ps1
```diff
@@ -0,0 +1,48 @@
+################################################################################
+##  File:  Validate-NSIS.ps1
+##  Team:  CI-X
+##  Desc:  Validate NSIS installation.
+################################################################################
+
+$SoftwareName = 'Nullsoft Install System (NSIS)'
+$regKey = gci HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\* | gp | ? { $_.DisplayName -eq 'Nullsoft Install System' }
```
brcrista: I wouldn't base the validation on a registry key since these are prone to move.  If it's not going in PATH, the installation directory should be known to us, so we can just check for the exe there.

I think we should also have a canary build for this.  I made one for Inno Setup where I used one of the example projects from their documentation.

images/win/scripts/Installers/Validate-NSIS.ps1
```diff
@@ -0,0 +1,48 @@
...
+    Write-Host "The $SoftwareName installation directory registry value is not set"
+    exit 1
+}
+else
+{
+    Write-Host "The $SoftwareName installation directory registry value is set to: $installDir"
+}
+
+$exeFilePath = Join-Path $installDir 'NSIS.exe'
+if (!(Test-Path $exeFilePath))
```
brcrista: switch these around: if (Test-Path) then success, else failure.  In general, positive conditions are easier to follow than negative ones

images/win/scripts/Installers/Validate-Perl.ps1
```diff
@@ -0,0 +1,24 @@
+################################################################################
+##  File:  Validate-Perl.ps1
+##  Team:  CI-X
+##  Desc:  Validate perl
+################################################################################
+
+if (Get-Command -Name 'perl')
+{
+    Write-Host "perl on path"
```
brcrista: we should get a canary build here too.  besides the interpreter being in PATH, we also want to make sure the standard libs work

dakale:
It looks like your editor may have changed the line endings for `images/win/vs2017-Server2016-Azure.json`

Id rather not check those changes in. I suspect (but dont know for sure) that unix line endings are checked in here. You should be able to check out the latest version of that file and reapply your changes

joshmgross:
> It looks like your editor may have changed the line endings for `images/win/vs2017-Server2016-Azure.json`
> 
> Id rather not check those changes in. I suspect (but dont know for sure) that unix line endings are checked in here. You should be able to check out the latest version of that file and reapply your changes

Oops, my bad. I'll fix this up

images/win/scripts/Installers/Validate-CloudFoundryCli.ps1
```diff
@@ -0,0 +1,28 @@
...
+{
+    Write-Host 'cf is not on path'
+    exit 1
+}
+
+if( $(cf version) -match  '\d+\.\d+\.\d+' )
+{
+   $version = $Matches[0]
+}
+# Adding description of the software to Markdown
```
brcrista: Getting the version belongs to this part of the logic, so this comment should move up

images/win/scripts/Installers/Install-CloudFoundryCli.ps1
```diff
@@ -0,0 +1,25 @@
+################################################################################
+##  File:  Install-CloudFoundryCli.ps1
+##  Team:  CI-X
+##  Desc:  Install Cloud Foundry CLI
+################################################################################
+
+Import-Module -Name ImageHelpers
+
+# Download the latest cf cli exe
+Invoke-WebRequest -UseBasicParsing -Uri "https://packages.cloudfoundry.org/stable?release=windows64-exe&source=github" -OutFile cf-cli.zip
```
brcrista: Will the [Install-EXE](https://github.com/Microsoft/azure-pipelines-image-generation/blob/9737f02f9139cf5df554354d3f279e3195534032/images/win/scripts/ImageHelpers/InstallHelpers.ps1#L44) helper work for this?

images/win/scripts/Installers/Validate-NSIS.ps1
```diff
@@ -0,0 +1,33 @@
...
+else
+{
+    Write-Host "$SoftwareName is not installed"
+    exit 1
+}
+
+# Adding description of the software to Markdown
+$ChocoList = $(choco list --local-only nsis) | Select-String -Pattern "nsis" | Select-Object -First 1
+
+if($ChocoList -Match "\d+\.\d+")
```
brcrista: nit: `if (`

images/win/scripts/Installers/Validate-OpenSSL.ps1
```diff
@@ -0,0 +1,28 @@
...
+{
+    Write-Host "openssl on path"
+}
+else
+{
+    Write-Host 'openssl is not on path'
+    exit 1
+}
+
+if( $(openssl version) -match  '\d+\.\d+\.\d+\w' )
```
brcrista: `if (`

images/win/scripts/Installers/Validate-GitVersion.ps1
```diff
@@ -0,0 +1,26 @@
+################################################################################
+##  File:  Validate-GitVersion.ps1
+##  Team:  CI-X
+##  Desc:  Validate GitVersion
+################################################################################
+
+$command = Get-Command -Name 'gitversion'
```
brcrista: Can you inline this in the `if ()` like the other scripts do?

images/win/scripts/Installers/Install-CloudFoundryCli.ps1
joshmgross: No, we're downloading a zip of the cf cli exe. Looks like Install-Exe downloads an exe and then runs that exe (I'm assuming some sort of installer). So its missing the unzip and its assuming that the exe is an installer rather than just the program we want.

images/win/scripts/Installers/Validate-GitVersion.ps1
joshmgross: I use the command, later on, to get the Version as well, so that's why it was put into a variable. I can inline it, but that would just be moving the $command variable down a few lines.

images/win/scripts/Installers/Install-NSIS.ps1
```diff
@@ -0,0 +1,11 @@
+################################################################################
+##  File:  Install-NSIS.ps1
+##  Team:  CI-X
+##  Desc:  Install NSIS
+################################################################################
+
+choco install nsis -y
+
+$NsisPath = "$env:ProgramFiles(x86)\NSIS\"
```
lokesh755: ${env:ProgramFiles(x86)} not $env:ProgramFiles(x86)

PR #313, opened 2018-10-19
Install Inno Setup

brcrista:
Inno Setup is a tool for making Windows installers.

Canary build: https://buildcanary.visualstudio.com/CanaryBuilds/_build/results?buildId=99992&view=logs


dakale:
Thanks Brian. This is looks mostly good to me. 

One question I have is with the canary build. Does ISCC.exe not get added to the path? Ideally builds would not need to know the fully qualified path to the exe. If that is not possible for any reason, a second best option would to be to at least report the path in the metadata file 

Is there an associated task that will know how to locate and run this? If so it would be good to also cover that usage in the canary build

brcrista:
@dakale sorry, I closed the PR before I saw this.  But that's a good point.
@DavidStaheli which do you prefer?

brcrista:
Chocolatey adds it to PATH, see https://buildcanary.visualstudio.com/CanaryBuilds/_build/results?buildId=100034&view=logs

I can add validation that it's in PATH in another PR.

brcrista:
Synced with @DavidStaheli offline and we decided to keep it in PATH

PR #325, opened 2018-10-23
Updated SSDT version to 15.8.0

lokesh755:


dakale:
Unblocks build failure for out of disk error

Confirmed that a good build has been ran

Issue #261, opened 2018-09-30
Add Inno Setup to Visual Studio 2017 Images


nick-thompson:
A free, popular tool for creating installers for Windows applications.
http://www.jrsoftware.org/isinfo.php

bryanmacfarlane:
@chrisrpatterson - I don't believe we're going to be adding (and supporting) software packages like this.

Your available options are to pull JIT (that agent runs as administrator).  You can even create a tool installer task and add it to the marketplace.

Run a private agent on an azure VM.

nick-thompson:
Thanks @bryanmacfarlane. So I've been playing with this for the better part of the past week, I don't mind installing it myself on a hosted agent (it's pretty simple with a `curl` and executing the installer with `/silent`), but the agent hangs currently because it's waiting for permission.

If I can run the `vs2017-win2016` agent with administrator privileges so as to avoid that permission prompt then I imagine the installation will be plenty easy (on a local Win10 installation I've been able to curl & install and everything works as expected after permission is given). I don't quite understand what you mean by "pull JIT" though? This is my config, how would I modify it to pull JIT?

```
- job: Windows
  pool:
    vmImage: 'vs2017-win2016'
  steps:
  - task: MSBuild@1
    inputs:
      solution: 'projects/**/*.sln'
      platform: 'x64'
      configuration: 'release'
      clean: true
  - task: ShellScript@2
    inputs:
      scriptPath: ./scripts/azure-build_Windows.sh
```

(My `azure-build_Windows.sh` script is a bash script that curls and tries to install InnoSetup, that's where my build hangs currently)

And my understanding with the tool installers was that this was a preview feature that, currently, seems unavailable to me?

bryanmacfarlane:
You would need to follow up with InnoSetup to find out why the installation hangs.  The agent on hosted runs with admin permissions.  JIT = Just in time which means install it as part of your build process which is what you're doing.

nick-thompson:
Ah, right, thanks.

I've asked around in other forums about what's going on here, and I'm getting the suggestion that this is a "UAC" popup and not an authorization popup (or therefore not the installer). Apparently that's evident because the popup message is "Do you want to allow InnoSetup to make changes to your device?" (I'm primarily an OS X user, so this is all relatively unfamiliar to me). Is there a way to disable UAC on the build agents?

nick-thompson:
[...]

PR #327, opened 2018-10-23
Validate that ISCC is in PATH

brcrista:
Feedback from #261


dakale:
LGTM, but just for sanity did you generate an image to check this script?

brcrista:
Yeah, I ran the build

PR #331, opened 2018-10-23
Update ISSUE_TEMPLATE.md

DavidStaheli:
This proposes updating the issue template to ask for:
1. The tool's license type (MIT, Apache, etc.)
1. The tool's disk space


DavidStaheli:
@vtbassmatt, I'm interested in your opinion if asking for these 2 new items will discourage people from making requests because of the work involved.

vtbassmatt:
License seems reasonable as a filtering function. We have real policy concerns with some licenses.

I'm not as sold on disk size -- won't we have to verify it anyhow, and if we have to verify it, why do we need people to tell us?

PR #335, opened 2018-10-25
Adjust Xcode list to include build number and path

max-lobanov:
Based on [this bug](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/51114?src=WorkItemMention&src-action=artifact_link) we should add build number and path for each Xcode


dakale:
Looks good. Is this ready to merge?

max-lobanov:
> Looks good. Is this ready to merge?

Yes

PR #338, opened 2018-10-25
Add java11 to the VS2017 images since java10 is removed

madhurig:
Orcale has removed Java10: https://www.oracle.com/technetwork/java/javase/downloads/jdk10-downloads-4416644.html, so we are updating to Java11 on the VS2017 VM images.


dakale:
@lokesh755 FYI. Not sure where the latest vs2017 image is at but this might need to go in

PR #353, opened 2018-11-05
Install WDK 10.0.17763.0

brcrista:
This requires the matching version of the Windows SDK (not currently available through the VS installer), so installing that as well.

**Testing**
- [x] Ran Packer build
- [x] Ran [canary build](https://buildcanary.visualstudio.com/CanaryBuilds/_build/results?buildId=106364&view=logs)


brcrista:
Blocked on #355

brcrista:
@joshmgross that's the idea.  There's a more general problem here though.  Currently we install a bunch of Windows SDK versions through the VS Installer.  However, the more recent versions have stopped being available through installer.  That's why we need to install the new version of the Windows SDK in this script.

PR #354, opened 2018-11-06
Fixing path error in installDir for mingw32-make.exe using choco install

tehho:
Validate mingw errored because the item did not exist.
Some figuring out later and this path poped out in the logs.


riezebosch:
Related to my #356 fix which achieves the same thing (but maybe more future proof).

PR #356, opened 2018-11-06
mingw: copy of mingw32-make.exe shim instead of hardlink to original executable

### LINKED ISSUE #355: Build failing because MinGW not installed on expected location
I don't know where this assumption comes from: https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Install-MinGW.ps1#L10

But when installing MinGW with the chocolatey command it gets installed in: 
````
Extracting C:\Users\packer\AppData\Local\Temp\chocolatey\mingw\8.1.0\mingwInstall.zip to C:\ProgramData\chocolatey\lib\mingw\tools\install...
````

So the [hardlink](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Install-MinGW.ps1#L15) fails on my build:
```
New-Item : Cannot find path 'C:\tools\mingw64\bin\mingw32-make.exe' because it does not exist.
```

And my [Validate-MinGW](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Validate-MinGW.ps1#L29) also fails:
```
2018-11-06T00:52:57.6595671Z     vhd: make is not on PATH
2018-11-06T00:52:57.6598536Z     vhd: Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
2018-11-06T00:52:57.6605723Z     vhd: At C:\Windows\Temp\script-5be0a1b6-6f3c-9179-5be6-0855e3153b7c.ps1:29 char:5
2018-11-06T00:52:57.6619020Z     vhd: + if (Get-Command -Name 'make')
2018-11-06T00:52:57.6627858Z     vhd: +     ~~~~~~~~~~~~~~~~~~~~~~~~
2018-11-06T00:52:57.6628576Z     vhd:     + CategoryInfo          : ObjectNotFound: (make:String) [Get-Command], CommandNotFoundException
2018-11-06T00:52:57.6636990Z     vhd:     + FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand
20
```

I see the [choco package](https://chocolatey.org/packages/mingw) is updated a week ago, ~but also in older packages I don't find any references to the `C:\tools` folder`~. 

In the previous packages the `Get-BinRoot` was used (which is replaced by [Get-ToolsLocation](https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/helpers/functions/Get-ToolsLocation.ps1#L33) and indeed that did install in `$env:ChocolateyToolsLocation` which normally points to `C:\tools`.

riezebosch:
* Chocolatey package no longer installs in tools directory
* Generated shims already discoverable
* Resolving shim and copy makes it future proof

closes #355


brcrista:
Since Chocolatey should add the install directory to `PATH`, using `where.exe` is a good solution to make this future-proof.

Why do you need to copy the file instead of hard linking it?  Since it is in the same directory it shouldn't matter but I am curious.

riezebosch:
It's a generated shim and not the actual executable. The shim only works for relative paths but a hardlink in the same folder would work.  Maybe Get-Command is even better than where.exe.

brcrista:
Ok.  Yeah I agree that `Get-Command` is the more PowerShell-y way to do it.  Though for this usage I don't think it makes a difference

riezebosch:
BTW: shims are a [chocolatey thing](https://chocolatey.org/docs/features-shim). I found it impossible to extract the target executable from the shim so I figured creating a copy of the shim with the desired name would suffice.

brcrista:
Thanks.  I'm building and testing your branch now.

brcrista:
Looks good.  Thanks!  🎉

PR #376, opened 2018-11-20
Go: Update to the latest revision of Go 1.11 on Linux

Caiyeon:
For linux images: changes go version from 1.11 to 1.11.2

Minor revisions 1.11.2 includes changes: 
https://golang.org/doc/devel/release.html#go1.11.minor
https://github.com/golang/go/issues?q=milestone%3AGo1.11.2

Which also includes changes for 1.11.1:
https://github.com/golang/go/issues?q=milestone%3AGo1.11.1

### DIFF
--- a/images/linux/scripts/installers/go.sh
+++ b/images/linux/scripts/installers/go.sh
@@ -32,4 +32,4 @@ function InstallGo () {
 # Install Go versions
 InstallGo 1.9 1.9.7 false
 InstallGo 1.10 1.10.4 false
-InstallGo 1.11 1.11 true
+InstallGo 1.11 1.11.2 true


images/linux/scripts/installers/go.sh
```diff
@@ -32,4 +32,4 @@ function InstallGo () {
 # Install Go versions
 InstallGo 1.9 1.9.7 false
 InstallGo 1.10 1.10.4 false
```
bishal-pdMSFT: The title says that Go1.10 is also updated, but it is still 1.10.4. Do you want to update it to 1.10.5?

bishal-pdMSFT:
This PR actually only updates linux image as opposed to the title which says windows as well. https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Install-Go.ps1 should also be updated for windows.

images/linux/scripts/installers/go.sh
Caiyeon: The title was edited by @DavidStaheli 

The original title was: `Update minor revision of go for 1.11`

I am indifferent to updating go 1.10

Caiyeon:
> This PR actually only updates linux image as opposed to the title which says windows as well. https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Install-Go.ps1 should also be updated for windows.

The title was updated by @DavidStaheli 

I am indifferent to updating go for windows images, but perhaps they should be in a separate PR?

I am only personally interested in linux images

images/linux/scripts/installers/go.sh
bishal-pdMSFT: Updated title accordingly

Issue #182, opened 2018-07-26
What is plan to support building container images based on different channel (sac, ltsc)


artisticcheese:
Is VSTS team working on a way to be able to build newer versions of container images (based off 1709 or 1803) using Windows VSTS hosted agent? Can may be a new type of agent be created with no VS installed and running off 1803 image where containers are used for builds and this way single agent can build ltsc and sac channels?

chrisrpatterson:
You should see the Hosted Windows Container pool in your account which is based Server 1803.

artisticcheese:
It fails to build inside it for 1709 images. 

```
[...]

PR #498, opened 2018-11-22
Change default VM size to a sku with nested virtualization

syedhassaanahmed:
Since both VM images support Docker, we should choose a default Azure VM sku which has [nested virtualization support](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/acu). Otherwise running docker tasks on these agents fails with error;
```
No hypervisor is present on this system
```


nelsonjchen:
What are the chances of this making it in? If not existing VM images, can this be done for `1809`-based OSes?

nelsonjchen:
The use case I'm trying to push forth in "https://github.com/rust-lang/docker-rust/issues/30" is building and testing images for all the differing in-between versions of Windows and apparently working around not facing some bugs in some older Windows's process isolation with VS Build Tools. For example, no online CI providers offer `1709` and we're having trouble installing VS Build Tools in an LTSC 2016 image without some process-isolation only call or filter driver stuff erroring out (All builds are on a building and test-only basis, but especially MSVC-based ones).

nelsonjchen:
Oof, https://github.com/Microsoft/azure-pipelines-image-generation/issues/182 , now I'm not so sure anymore. Here's to hoping the Windows Server 2019 image runs on a nested virtualization capable host.

syedhassaanahmed:
@lokesh755 @DavidStaheli @chrisrpatterson Can you please take a look at this PR? Thanks

PR #503, opened 2018-11-27
Add names of basic CLI tools to README

brcrista:
I ended up not adding --version of any of the tools to avoid having to parse the version output.


dakale:
LGTM. I may in the future rewrite this to be a little more clear. libcurl isnt a cli for example, and in general these only have in common the fact that they are commonly used packages. Some happen to bring cli tools like jq but the lib* packages, while worth mentioning, arent cli. I think this metadata is more to inform people about what to expect rather than a full catalog. Going back to libcurl, I dont think anyone will actually read the entire metadata doc before running apt install libcurl, they will just install libcurl and since its already installed, the operation will be a nop. 

Consider this comment for posterity rather than direct feedback about the PR. It just illuminates the fact the metadata document isnt really clear in its purpose, today.

PR #535, opened 2018-12-12
Include perl6 installation script for Linux

kdbeall:
Taking a stab at this :) I saw that perl was part of the VS 2017 image, so I thought adding it to Linux would be a good idea.

### DIFF
--- a/images/linux/scripts/installers/perl6.sh
+++ b/images/linux/scripts/installers/perl6.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+################################################################################
+##  File:  perl6.sh
+##  Team:  CI-Platform
+##  Desc:  Installs Perl6
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install Perl
+wget https://rakudo.perl6.org/downloads/star/rakudo-star-2018.10.tar.gz
+tar xfz rakudo-star-2018.10.tar.gz
+cd rakudo-star-2018.10
+perl Configure.pl --gen-moar --make-install --prefix ~/rakudo
+
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! command -v perl6; then
+    exit 1
+fi
+
+# Document what was added to the image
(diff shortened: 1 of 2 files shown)


kdbeall:
@dakale I realized this PR is a bit premature! Sorry about that. I need perl6

kdbeall:
@dakale This should be ready now!

dakale:
I tested this on Ubuntu 16, which is what our Ubuntu images are based on and it seems like `perl6` is not the correct package name. Did you try this on Ubuntu 16 or some other system?

kdbeall:
@dakale That makes sense. I'm using Ubuntu 1804! I'll fire up vagrant.

kdbeall:
@dakale Because perl6 isn't available on Ubuntu 1604, the standard way is to get the source and then compile. Is compiling from source acceptable?

dakale:
@kdbeall It seems like the way we are going with these images, it is hard to continue adding every new tool. Part of the strategy going forward is [job containers](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=vsts&tabs=yaml)

Basically this will let you bring a container, and the job will run in this container rather than directly on the hosted. This lets you have more flexibility in bringing the tools you need, and keeping the tools up-to-date (just pull a latest image, rather than wait for host image generation updates)

I did a quick little sample to prove this out before recommending you to try it. 

I made a simple yaml definition as follows:
```yml
container: rakudo-star:latest
steps:
  - script: | 
      perl6 -e "say 'Hello'"
```
This pulls the latest rakudo-star image (from https://hub.docker.com/_/rakudo-star/) and executed the steps inside the container instead of on the host. And of course, theres more information about how to use container in general in the docs I linked to. 

Do you think this is something that would suit your needs? I think its the better long term solution than maintaining a perl6 installation on the host image, since its not an often asked for tool

DavidStaheli:
Closing this because it seems a decision was made to not add Perl 6 to the 16.04 image.  Otherwise, please re-open it.  Thanks for the contribution, @kdbeall!

Issue #413, opened 2018-11-21
Java: Update to Gradle 4.10.2 (or latest) (Windows)


DavidStaheli:
Java: Update to Gradle 4.10.2 (or latest) (Windows)

Issue #385, opened 2018-11-21
Android: Add Android SDK Build-Tools 28.0.3 (Linux, Windows)


DavidStaheli:
Android: Add Android SDK Build-Tools 28.0.3 (Linux, Windows)

acortelyou:
+1

ludovic-garcia:
+1

JesusSpiegel:
A problem occurred configuring project ':app'.
Failed to install the following SDK components:
build-tools;28.0.3 Android SDK Build-Tools 28.0.3
The SDK directory is not writable (/usr/local/lib/android/sdk)

nguyenlamlll:
Hi everyone, is there any progress on this issue?

I'm encountering this issue on Linux agent only, meaning build steps on Windows works fine. I opened a question [here](https://stackoverflow.com/questions/53870742/sdk-directory-is-not-writable-when-building-android-project-on-ubuntu-agent-of-a) in case someone needs more details.

On Windows, this works:
```
.\gradlew assembleDebug
```

On Linux (Hosted Ubuntu 1604), I'm using:
```
chmod 775 gradlew
./gradlew assembleDebug
``` 
and the build throws this exception:
```
* What went wrong:
A problem occurred configuring project ':app'.
> Failed to install the following SDK components:
      build-tools;28.0.3 Android SDK Build-Tools 28.0.3
  The SDK directory is not writable (/usr/local/lib/android/sdk)
```

DavidStaheli:
@nguyenlamlll the 28.0.3 build tools should be available on hosted Linux in the next 3 weeks.  We're changing our processes to make rollout much faster than this going forward.

Issue #383, opened 2018-11-21
Android: Add Android SDK Build-Tools 28.0.1 (Linux, Windows)


DavidStaheli:
Android: Add Android SDK Build-Tools 28.0.1 (Linux, Windows)

PR #607, opened 2018-12-12
Add/update Android and Java tool versions

DavidStaheli:
This PR addresses Android Build-Tools versions needed on the Linux and Windows images.  This matches the macOS image and allows anyone to bring their Gradle file [which can rely on any version](https://developer.android.com/studio/releases/build-tools) of these build tools.
#383, #384, #385

This PR also updates versions of Gradle and Maven:
#357, #413 

It also updates the Windows image to use Java 1-8.0.**192** instead of 1-8.0.**181**.


images/linux/scripts/installers/android.sh
```diff
@@ -38,18 +38,32 @@ echo "y" | ${ANDROID_ROOT}/tools/bin/sdkmanager --sdk_root=${ANDROID_SDK_ROOT} \
...
     "build-tools;24.0.3" \
+    "build-tools;24.0.2" \
+    "build-tools;24.0.1" \
+    "build-tools;24.0.0" \
     "build-tools;23.0.3" \
+    "build-tools;23.0.2" \
     "build-tools;23.0.1" \
     "build-tools;22.0.1" \
     "build-tools;21.1.2" \
+    "build-tools;20.0.0" \
```
madhurig: Do we need to add older versions that we didn't have before?

images/linux/scripts/installers/java-tools.sh
```diff
@@ -30,14 +30,14 @@ apt-fast install -y --no-install-recommends ant ant-optional
...
 unzip -d /usr/share maven.zip
 rm maven.zip
-ln -s /usr/share/apache-maven-3.5.4/bin/mvn /usr/bin/mvn
-echo "M2_HOME=/usr/share/apache-maven-3.5.4" | tee -a /etc/environment
+ln -s /usr/share/apache-maven-3.6.0/bin/mvn /usr/bin/mvn
+echo "M2_HOME=/usr/share/apache-maven-3.6.0" | tee -a /etc/environment
 
 # Install Gradle
-gradle_version="4.10"
+gradle_version="5.0"
```
madhurig: Any compat issues with Gralde 5 that we should test for?

images/linux/scripts/installers/android.sh
DavidStaheli: We won't remove them.  We need them for people who bring gradle files that depend on these build tools.  But they aren't a whole SDK.  This is the pattern that App Center established for their Android build tool versions.

images/linux/scripts/installers/java-tools.sh
DavidStaheli: None that I see in Gradle 5.0 release notes, for Android or otherwise.

images/linux/scripts/installers/android.sh
```diff
@@ -38,18 +38,32 @@ echo "y" | ${ANDROID_ROOT}/tools/bin/sdkmanager --sdk_root=${ANDROID_SDK_ROOT} \
     "platforms;android-17" \
     "platforms;android-15" \
     "platforms;android-10" \
+    "build-tools;28.0.3" \
```
lokesh755: Corresponding "DocumentInstalledItem" need to be added for new versions.

DavidStaheli: Done - thanks for catching that.

PR #618, opened 2018-12-20
Fix Docker image manifest references

heaths:
Not only correct syntax, but allows for easy copy-and-paste into Dockerfiles from this document.


lokesh755:
Updated the Readme in different PR https://github.com/Microsoft/azure-pipelines-image-generation/pull/619

PR #619, opened 2018-12-20
Releases/1809

lokesh755:


heaths:
That is not the right manifest digest format for image references. Please see my PR again for the proper syntax. It's also documented on Docker's web site.

PR #620, opened 2018-12-20
Correct the image digest manifest format

heaths:
See https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier


lokesh755:
We generate the markdown file, so we don't update it directly. I've updated in the powershell script to format the digest https://github.com/Microsoft/azure-pipelines-image-generation/pull/625

Issue #394, opened 2018-11-21
Version control: Update to Mercurial 4.8 (or latest) (Linux)


DavidStaheli:
Version control: Update to Mercurial 4.8 (or latest) (Linux, macOS)

DavidStaheli:
The [latest available](https://www.mercurial-scm.org/wiki/Download) version for Ubuntu Xenial is 4.4.1.  The latest can be installed with:
```
add-apt-repository ppa:mercurial-ppa/releases -y
apt-get update
apt-get install -y mercurial
```

DavidStaheli:
Completed with pull request #648.

PR #648, opened 2019-01-18
Install the latest Mercurial on Linux (currently 4.4.1)

DavidStaheli:
This addresses issue #394 by installing the latest version of Mercurial from a different repository, since the version being installed to date is old.


DavidStaheli:
Mercurial canary tests have been added for Linux and VS2017.

Issue #495, opened 2018-11-21
Languages: Add Rust 1.34.0 (or latest) (Linux, Windows)


DavidStaheli:
Languages: Rust (Linux, macOS, Windows)

spontoreau:
Sounds great 👍

PR #653, opened 2019-01-21
Add Rust to Linux and Windows

DavidStaheli:
This addresses issue #495.
Canary tests are created in the Linux and VS2017 folders.

### DIFF
--- a/images/linux/scripts/installers/rust.sh
+++ b/images/linux/scripts/installers/rust.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+################################################################################
+##  File:  rust.sh
+##  Team:  CI-Platform
+##  Desc:  Installs Rust and its common tools
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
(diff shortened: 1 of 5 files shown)


calebcartwright:
Just want to say that I am pumped about seeing this! The rustup installs are adding at least a minute to my builds so this will be great to have. If you do need to end up going a different route than rustup, I would kindly request that [clippy](https://github.com/rust-lang/rust-clippy) be in the list of tools for inclusion. 

All the relevant docs indicate that [clippy installation](https://github.com/rust-lang/rust-clippy#step-2-install-clippy) be done via `rustup component add` as installation from crates.io is deprecated (see https://crates.io/crates/clippy)

calebcartwright:
The Rust team uses Azure Pipelines for Docs.rs (https://github.com/rust-lang/docs.rs). Sharing in case their Azure Pipeline configs are helpful

DavidStaheli:
Thanks very much, @calebcartwright!  I'm hoping to get back to this PR ASAP and your link is a helpful reference.

DavidStaheli:
Closing this because it was addressed by #900 .

Issue #145, opened 2018-05-19
.NET Core 2.1 Global Tools need location in the path


onovotny:
.NET Core Global Tools need a location added into the path Windows, Linux, and macOS.

See here for details:
https://natemcmaster.com/blog/2018/05/12/dotnet-global-tools/#path-and-command-not-after-installing

lokesh755:
I could see  "$env:USERPROFILE/.dotnet/tools" is getting added for Windows, Did you see it missing for Linux or macOS ?

bryanmacfarlane:
@kmkumaran

kmkumaran:
@sachinma can you take a look at this.

DavidStaheli:
This appears to be fixed by [Microsoft/azure-pipelines-tasks#8432](https://github.com/Microsoft/azure-pipelines-tasks/pull/8432).  Can this issue be closed?

sachinma:
Adding @bishal-pdMSFT

bishal-pdMSFT:
https://github.com/Microsoft/azure-pipelines-tasks/pull/8432 fixes this problem only when .Net Core tool installer task is used. It does not fix the issue in hosted images.

bishal-pdMSFT:
@DavidStaheli can you please add me to this repo so that I can assign the issue to myself and possibly raise PR.

DavidStaheli:
@bishal-pdMSFT you should receive an email invite - thanks!

bishal-pdMSFT:
I don't think we can set Global tools path for hosted image. Global tools are [user-specific](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-global-tool). Any user specific path set in image generation scripts will be removed by Sysprep as it deprovisions the user. Also the agents run with a different user from the one Packer uses.

bishal-pdMSFT:
> Also the agents run with a different user from the one Packer uses.
I might be wrong about this though. @lokesh755 does packer run with same user as hosted agent?

bishal-pdMSFT:
@onovotny @scottaddie Given that this issue can't be fixed in hosted image, do you think an alternative would be to fix this in ".Net Core" task. This task will additionally add global tools path to PATH(for that release). By default this option will be turned off, but if a user chooses to use this task to install global tool, they can turn it on.

hiyadav:
The Fix for this issue is merged and should roll-out with the next agent image update.
#670

PR #670, opened 2019-01-30
Added dotnet tool global path to VS2017 and ubuntu image global path variables

hiyadav:
This PR attends to the ongoing issue for adding support for .Net global tool path availability on Hosted 2017 and Ubuntu images : #145

### DIFF
--- a/images/win/scripts/Installers/Install-DotnetSDK.ps1
+++ b/images/win/scripts/Installers/Install-DotnetSDK.ps1
@@ -51,3 +51,4 @@ $dotnetReleases | ForEach-Object {
 }
 
 Add-MachinePathItem "C:\Program Files\dotnet"
+Add-MachinePathItem "C:\Users\VssAdmisnitrator\.dotnet\tools"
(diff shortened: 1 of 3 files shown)


images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,4 +50,5 @@ $dotnetReleases | ForEach-Object {
     }
 }
 
-Add-MachinePathItem "C:\Program Files\dotnet"
\ No newline at end of file
+Add-MachinePathItem "C:\Program Files\dotnet"
+Add-MachinePathItem "C:\Users\VssAdmisnitrator\.dotnet\tools"
```
chrisrpatterson: Should this be %USERPROFILE%\.dotnet\tools?  That way it won't be affected if we choose to run the build as a different user.

hiyadav: Agreed, I just was not sure if doing that at image generation time might resolve %USERPROFILE% to packer build user when adding the path and might not add the new user path when a new user is created to run agent.
Will test this out though and change here, if it works out.

hiyadav: Using variables is not working as the "%USERPROFILE%" is not getting expanded while setting and retrieving the PATH env variable. Hence using the hard coded path of VssAdministrator

shadyysf:
@hiyadav , we should add verification scripts to make sure the install scripts are doing what they're expected to do. We should also make sure to have a Canary build that runs a regression test for this scenario.

hiyadav:
@shadyysf i have added validation step for .Net global tool path in validation script for windows and updated all .Net Core canary build to validate .Net core tools path is set in agent.

PR #673, opened 2019-01-31
Moby-engine installation

ds-ms:
This PR replaces docker-ce installation with ms-moby installation.


images/linux/scripts/installers/docker-moby.sh
```diff
@@ -19,10 +17,10 @@ if ! IsInstalled $docker_package; then
     apt-get remove -y docker docker-engine docker.io
```
bansalaseem: Should we remove the moby engine here?

images/linux/scripts/installers/docker.sh
```diff
@@ -21,10 +21,10 @@ if ! IsInstalled $DOCKER_PACKAGE; then
     apt-get remove -y docker docker-engine docker.io
```
bansalaseem: Same as above

images/linux/scripts/installers/docker-moby.sh
ds-ms: @bansalaseem Done.

images/linux/scripts/installers/docker.sh
ds-ms: done

images/linux/scripts/installers/docker-moby.sh
```diff
@@ -8,21 +8,20 @@
...
     apt-get update
     apt-get install -y apt-transport-https ca-certificates curl software-properties-common
-    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
-    add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $LSB_CODENAME stable"
+    curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
+    curl https://packages.microsoft.com/config/ubuntu/16.04/prod.list > /etc/apt/sources.list.d/microsoft-prod.list
     apt-get update
-    apt-get install -y docker-ce
+    apt-get install -y moby-engine moby-cli
+    usermod -a -G docker $USER
```
dakale: This runs in Packer as the `packer` user which is deleted at the end of provisioning, so this is not really needed

ds-ms: Removed

Issue #448, opened 2018-11-21
Deployment: Add Terraform 0.11.10 (or latest) (Windows)


davidstaheli:
Deployment: Add Terraform 0.11.10 (or latest) (macOS, Windows)

DavidStaheli:
See this pull request: #679

PR #679, opened 2019-02-01
Add HashiCorp tools install scripts for both Windows and Linux(Ubuntu)

mortenlerudjordet:
I have written scripts to cover installation of HashiCorp tools (Consul, Packer, Terraform (only win) and Vault).

For Linux (Ubuntu) I have made the script generic, and with only a change of a variable the different tools are targeted. The script also supports setting a specific version to "install", or just use "latest" to retrieve the newest one.
Another note is that the scripts update the PATH variable in both /etc/environment and create a bash script under /etc/profile,d/ called setpath.sh. This adds the same path content, but in a export path="updated path" format.

I have tried to follow the examples set by previous scripts in this repo.


DavidStaheli:
This fixes #448.

JustinGrote:
LGTM!

ryanspletzer:
I would love to see this one merged. :) I use Packer and Vault today and have to arrange alternate means to get them onto the build hosts.

mortenlerudjordet:
If nobody wants this PR, I will just close it

gep13:
@kmkumaran @RoopeshNair @davidstaheli @kaylangan  It would be great to get some eyes on this, and hopefully a merge.  Is there something preventing this PR from getting merged?  It would be great to get some feedback on it.

elduddz:
How many people is more than nobody? 😁

mortenlerudjordet:
Wait, is this a trick question? 

Admittedly imprecise wording on my part, though it was more of a hands up please decide one way or the other. Wanted to do some autumn cleaning on my repos, and would be great if I could close out my fork. :do_not_litter:

bryanmacfarlane:
Have you considered making an installer task instead?  (see the node installer task in azure-pipelines-tasks)

These seem like they are easy to download.  The typical choice boils down to the size of the tools (whether it's practical to install on the fly).

mortenlerudjordet:
I'm generally against installing software as part of pipeline execution. Just because it will add a significant time overhead. IMO it is way better to bake these into the image/container used for build/release logic to run on.

elduddz:
I've thumbed it up a long time 🤪

would be useful. I've been adding terraform (and removing all the validation 😁) to my local run.

bryanmacfarlane:
The hosted pools are running on the back bone.  When you say significant delays, do you have numbers?  On the counter side we also have a very real disk budget so we’re weighing that against the practicality of pulling JIT.  For example I believe node is bigger than these tools and it’s pulled very quickly 

Please size and  include times from hosted pools

mortenlerudjordet:
@bryanmacfarlane, don't have hard numbers for you. Though in our experience installing a couple tools does not boost the delay numbers (admittedly "significant" was a bad chosen word on my part). Though doing 10 or even more had impact enough for us that we pursued building a pipeline for creating the images with the tools we need prebaked. Instead of doing it as part of the build pipeline each time.

Now milage as with everything may vary for others. If the guidance is to do these things through tasks instead, fair enough. I just wanted to give something back as we used this repo as our starting point. Though if this is not the right type of content, a few words in this PR would have been enough for me to have closed it a long time ago.

bryanmacfarlane:
I’ll defer to @kmkumaran since he owns this area and will maintain.  Just wanted to detail the trade offs.

kmkumaran:
For Terraform, we are considering the tool installer path than baking that in the hosted images. + @RoopeshNair 

The tool installer is currently available in the [marketplace extension](https://marketplace.visualstudio.com/items?itemName=ms-devlabs.custom-terraform-tasks).

mortenlerudjordet:
Then I'll close this PR.

PR #680, opened 2019-02-01
change installer to use PowerShell Team's installer

SteveL-MSFT:
Based on community feedback, request to use a Microsoft authored way to install PSCore6.  This uses the installer script written and maintained by PowerShell Team.

### DIFF
--- a/images/win/scripts/Installers/Install-PowershellCore.ps1
+++ b/images/win/scripts/Installers/Install-PowershellCore.ps1
@@ -4,4 +4,4 @@
 ##  Desc:  Install Powershel Core
 ################################################################################
 
-choco install powershell-core --version 6.1.1 -y
+Invoke-Expression "& { $(Invoke-RestMethod https://aka.ms/install-powershell.ps1) } -UseMSI -Quiet"


ferventcoder:
This begs the question - doesn't it make sense for the PowerShell team to own the PowerShell Core Chocolatey package?

gep13:
@SteveL-MSFT are you able to share the community feedback that was received regarding this installation method?

markekraus:
@gep13 The PowerShell Core package on Chocolately is maintained by a 3rd party and is not owned or maintained by MSFT nor the PowerShell team. The current is tall method does not check the authenticity of the contents of the package. It is thus possible for a third party to modify the package and replace it with their own code. As Azure DevOps is a MSFT product and PowerShell Core is a MSFT product, there is a reasonable expectation that MSFT will maintain the chain of trust between the 2.

gep13:
@markekraus I am slightly confused by this stance, and I would like to delve into this further...

> The current is tall method does not check the authenticity of the contents of the package.

Depending on exactly what you are referring to, this simply isn't true.  The Chocolatey Package includes a checksum for both the x86 and x64 msi's, which Chocolatey will validate at the point of installation.  If these checksums do not match what is downloaded, Chocolatey will halt the installation.  The Package also contains a VERIFICATION.txt which explains how these checksums can be independently verified.

>  It is thus possible for a third party to modify the package and replace it with their own code

The package is maintained by a very well respected Community Package Maintainer in the form of @darwinjs, who maintains both the PowerShell (2,294,234 downloads) and PowerShell Core (60,808 downloads) packages.  In addition to this, each package goes through a rigorous moderation process, so the likelihood of this occurring is highly unlikely.

Further, Chocolatey is installed by default on the Azure Pipeline base image, and it is currently used to install 20 packages (some of which are also software applications that come from Microsoft):

image

And there are open PR's for adding more packages using Chocolatey.  Microsoft officially recognised Chocolatey, and indeed Boxstarter, at the Microsoft Build Conference in 2018 when they announced [this repository](https://github.com/Microsoft/windows-dev-box-setup-scripts) which shows how to setup different development environments using both of these tools.  

So, a genuine question, if Chocolatey can be trusted to install these other packages, why is it different for the PowerShell Core package?

As mentioned by @ferventcoder, ideally the PowerShell Team would become the maintainers of the PowerShell and PowerShell Core packages on Chocolatey, which I think would address most of your concerns.  We would certainly be more than happy to add people from the PowerShell Team as maintainers of these packages, as from our point of view, this is the ideal case, when the Software Authors are also the package maintainers.

/cc @yodurr @bitcrazed

ferventcoder:
@ryanspletzer ^^

kilasuit:
@ferventcoder I've asked them to maintain it in this issue https://github.com/PowerShell/PowerShell/issues/2708 

however this raises a further question of should MSFT have their own Chocloatey feed as per the comment I made here https://github.com/PowerShell/PowerShell/issues/8663#issuecomment-456272838

IMO the right way forward would be to have either a MSFT backed Chocolatey feed where all Microsoft product teams release a Choco package that is trusted end to end for installation of any and all Microsoft software and would become the defacto place for installing items like VS, VSCode, Pwsh, Office etc etc.

@gep13 have a look into this issue https://github.com/PowerShell/PowerShell/issues/8663 as I believe this and all the linked ones was the push behind having a Microsoft supported install method

ferventcoder:
It is also worth noting that Darwin received a community award from the PowerShell team at PowerShell Summit 2017 for his work on the PowerShell and SSH packages - based on that I would say he is a trusted member of both the Chocolatey AND the PowerShell community.

gep13:
@markekraus I also just thought of something else...

Are Microsoft, or the PowerShell Team, responsible for the apt-get package that is being used here:

https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/scripts/installers/powershellcore.sh#L14

If so, is there a reason why they don't also support one of the main package manager's for Windows as well?

markekraus:
My concern with this is not chocolatey. My concern is the 3rd party package being used to install 1st party software. It's a break in trust, even if Darwin is a stand-up dude and the community (of which I'm a part of) and PowerShell Team (with whom I often work with closely) think so. My concern is also not about Darwin maintaining a PowerShell Core package (this is not the appropriate forum to bring that up).

Let's frame it this way, in the event that Darwin turns to the dark side or, (perhaps way more likely but still an edge case) his chocolatey account is compromised, Is Microsoft going to take responsibility? It's MSFT software on a MSFT service. I would completely expect MSFT to be liable, but using this 3rd party package for their own software muddies the waters.

I realize this gets into a "turtles all the way down" slippery slope (e.g. "What is chocolatey is compromised?"), but, we _have_ a MSFT provided way to install PowerShell Core and in such a scenario, it is favorable W.R.T. trust to use it over a 3rd party package.

If this project is going to use choco to install a 3rd party package of 1st party software, it should at least validate it first. And, no, you cannot rely on the package and chocolatey to do that for you as you cannot off load the trust onto the  object of contention. You would need to add that validation here (checking the code signing cert and failing the image build, for example).

Ideally, yea, I would like the PS Team to own the choco package. To answer your question about `apt` (and other Linux packages of PS Core) I believe those are in fact published by MSFT. I believe that pretty much every major package feed with the exception of choco has an official MSFT package published to it by MSFT.

Also, let's not assume that I _do_ trust the other packages just because I haven't opened issues for them. IMO, if the package is published and maintained by the actual company/author that creates and maintains the software in the package, then it is not a trust concern. But, for any package where the package publisher is not the same, there should be some additional checks in place. PowerShell Core is just the 1st I become aware of this potential trust issue. 

I have a security department that is very detailed about its threat profiles for the technologies we use for production resources. This is exactly the kind of thing they would flag for remediation or ban us from using. 😢

gep13:
@markekraus regarding the trust issues that you bring up, I totally agree.  In fact, this is documented at length on the website:

https://chocolatey.org/docs/security#organizational-use-of-chocolatey

Where we actually recommend against the use of the Community Package Repository when using Chocolatey in an Organisational context.  Instead, your own, vetted list of packages would be maintained, that had been through the verification process that you mention above.  Using packages directly from the Community Package Repository is indeed a risk, and one which each individual, or organisation, need to be comfortable with.

DarwinJS:
I keep raising the issue of an official Microsoft feed as I think it would be great.
Also I enjoy the challenge of doing complex packaging for something that will run everywhere and why I pick up such packages - but if a vendor wants to take up packaging - I'm cool with that.  

Unlike OpenSSH this package is much more of a simple shell for the underlying MSI.

However, as of the ccleaner hack there is still caution to be had that a vendor's packages for their own software are automatically trustworthy.  Hackers are becoming aware of leveraging automated CI / CD to piggy back in the packaging of a vendor or to inject into software before it is signed by the company who made it.

Presume Breach,
D.

SteveL-MSFT:
Regarding an official Microsoft chocolatey feed, this was discussed by the PowerShell team and I thought we agreed it was a good idea.  This would be owned by @edyoung who owns the PowerShell Gallery.  However, it's not in any near term plans.  cc @SydneyhSmith who is the PM for this.

gep13:
@SteveL-MSFT @edyoung @SydneyhSmith if there is anything that we, the @chocolatey team, can do to help with setting up this Chocolatey Feed, then please feel free to reach out.  We would be happy to assist where we can.

joeyaiello:
Just wanted to echo @markekraus above. @darwin-msft has done phenomenal work in maintaining the Chocolatey package, but this is about building a trust chain that's Microsoft all the way down. Without ruling out our maintenance of the PS Core package on the community feed, it unfortunately wouldn't solve the problem being addressed here. (Running our own feed *would*, which is what we do with apt/yum/etc at https://packages.microsoft.com/ but as @SteveL-MSFT said, that's not on our roadmap at the moment.)

Issue #242, opened 2018-09-17
C/C++: Add vcpkg 0.0.113 (or latest) (Linux, Windows)


qmfrederik:
I'm trying to migrate some of my open source projects from AppVeyor + Travis to Azure pipelines.

I've noticed [VCPKG](https://github.com/microsoft/vcpkg) is not installed by default on the Azure Pipeline images for VS.

It's installed by default on AppVeyor, hence I'm checking in to see whether there's a possibility of having it available in the Azure Pipeline images as well?

Installing VCPKG takes a while, and it may make sense to pre-compile some common librarys (say OpenSSL) as well.

DominicCabral:
Any update on this?

vSzemkel:
I'm using AppVeyor. They have a vcpkg preinstalled on build machines for quite a long time.
C'mon guys, this is MSFT orginated package manager. Add the support for this to Azure Pipelines, please.

qmfrederik:
As a sidenode - it would make sense to pre-install some commonly used packages which take a while to compile. OpenSSL comes to mind; having it pre-installed in its x86 and x64 configuration could easily shave a couple of minutes off every build.

DavidStaheli:
Work on this has begun.  :clap:

DavidStaheli:
## Thank you, @vmapetr! 🥇

SeanFarrow:
Is this available in the hosted windows image as yet?
Were any packages pre-installed in to the image?

kaylangan:
This is available on the [Ubuntu 16.04](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md) and the [VS2017](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2017-Server2016-Readme.md#vcpkg) images. It is currently rolling out on the [VS2019](https://github.com/microsoft/azure-pipelines-image-generation/blob/releases/VS2019/151.5/images/win/Vs2019-Server2019-Readme.md#vcpkg) image.

SeanFarrow:
Hi,

Do we have an ETA as to when this will be available on the VS2019 images?
Also, are there plans to make this available on the Mac Osx images?
Thanks,
Sean.

kaylangan:
> Do we have an ETA as to when this will be available on the VS2019 images?

Depends on which ring your account is on, but were currently on track for EOW for all rings.

[...]

PR #682, opened 2019-02-04
C/C++: Add vcpkg 0.0.113 (or latest) (Linux, Windows)

vmapetr:
This PR fixes https://github.com/Microsoft/azure-pipelines-image-generation/issues/242 Adds vcpkg 0.0.113 (or latest) (Linux, Windows):
* Implemented script to install vcpkg on Windows
* Implemented script to install vcpkg on Linux
* Implemented separate installation script for GNU Cpp Compiler (g++) which is required for vcpkg on Linux.

**For discussion**
* Should some of the vcpkg packages to be pre-installed as per one of the issue comments?

**Tests**
* Run scripts manually on local boxes (Windows 10, Ubuntu 16.04)
* Via Azure VM provisioning and then verifying vcpckg (`vcpkg install openssl`)


sgrebnov:
@vijayma, @azureDaveOps, @madhurig, @vidStaheli - could you please review this PR

images/linux/scripts/installers/gcc.sh
```diff
@@ -0,0 +1,26 @@
...
+################################################################################
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install GNU C++ compiler
+add-apt-repository ppa:ubuntu-toolchain-r/test -y
+apt-get update -y
+apt-get install g++-7 -y
+
```
DavidStaheli: The build-essential package is already installed [here](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/scripts/installers/build-essential.sh) and includes gcc 5, so this installation of gcc 7 is a welcome improvement.

mitza-oci:
How should a project using the pre-installed `vcpkg` discover the location of the `installed` directory for use in the later build steps?  We could search the path for `vcpkg` but it seems like this shouldn't be needed in each project.

Our current azure-pipelines.yml for reference: https://github.com/objectcomputing/OpenDDS/blob/38e578946bdfeb75c3ee266528decf934ff29a49/azure-pipelines.yml#L22

sgrebnov:
@mitza-oci - Please confirm the ideal scenario for you to have [VCPKG_ROOT](https://github.com/Microsoft/vcpkg/blob/1d5e22919fcfeba3fe513248e73395c42ac18ae4/docs/users/config-environment.md#vcpkg_root) env variable to be pre-defined on VM?

PS.  Similar to appveyor (https://www.appveyor.com/docs/lang/cpp/#vc-packaging-tool) I think it is also safe to consider vcpkg is pre-installed to a specific folder ('C:\vcpkg' on Windows and '/usr/local/share/vcpkg' on Linux).

mitza-oci:
Yes, sounds like that would work.

ras0219-msft:
This will cause problems for users that are currently cloning their own vcpkg instance -- when they build their own vcpkg executable and run it, the `VCPKG_ROOT` environment variable will override their local paths with the central one in the image.

This is undesirable because the model of vcpkg is that the VCPKG_ROOT location stores the version information for what libraries to install. If a user has cloned their own copy, they presumably wish to install those version ;).

We would instead recommend that users either clone their own copy of vcpkg or hardcode the path for the azure pipelines images.

max-lobanov:
Hi,

It looks like these changes have overridden the default `VCPKG_ROOT` vcpkg variable that should point to the `$HOME/.vcpkg`. It causes permission issues for our customers on Linux images
To fix the issue we need to rename `VCPKG_ROOT` variable to something like `VCPKG_INSTALLATION_ROOT`. So `VCPKG_INSTALLATION_ROOT` will point to the vcpkg installation dir and `VCPKG_ROOT` will point to the `$HOME/.vcpkg` by default
As a temporary workaround customers need to override `VCPKG_ROOT` to `$HOME/.vcpkg` in their scripts before using  `vcpkg`.

@DavidStaheli , I think we need to log bug for this issue and fix it. Please confirm

cc: @ras0219-msft, @sgrebnov

DavidStaheli:
I created issue #769 to fix this.

PR #683, opened 2019-02-04
Enable short name support priot to installing VS on Windows VMs

madhurig:
Fix for the issue reported here: https://developercommunity.visualstudio.com/content/problem/431025/xamarin-android-aot-complition-problem.html

We have determined there is a change in the Azure VM settings (unsure if this was a policy change in Azure or a windows update change)

### DIFF
--- a/images/win/scripts/Installers/Vs2017/Install-VS2017.ps1
+++ b/images/win/scripts/Installers/Vs2017/Install-VS2017.ps1
@@ -17,6 +17,16 @@ Function InstallVS
 
   try
   {
+    Write-Host "Enable short name support on Windows needed for Xamarin Android AOT, defaults appear to have been changed in Azure VMs"
+    $shortNameEnableProcess = Start-Process -FilePath fsutil.exe -ArgumentList ('8dot3name', 'set', '0') -Wait -PassThru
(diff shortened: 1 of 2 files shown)


images/win/scripts/Installers/Vs2017/Install-VS2017.ps1
```diff
@@ -17,6 +17,16 @@ Function InstallVS
 
   try
   {
+    Write-Host "Enable short name support on Windows needed for Xamarin Android AOT, defaults appear to have been changed in Azure VMs"
+    $shortNameEnableProcess = Start-Process -FilePath fsutil.exe -ArgumentList ('8dot3name', 'set', '0') -Wait -PassThru
+    $shortNameEnableExitCode = $shortNameEnableProcess.ExitCode
+
+    if ($shortNameEnableExitCode -ne 0)
```
joshmgross: Do we ever expect this to fail?

madhurig: It might, I don't know for sure. The time I have seen it fail is invalid arguments.

madhurig: We want to stop if it does fail to figure out what caused the issue.

PR #690, opened 2019-02-07
Adding support for WinAppDriver to Windows 10 1809 image

hassanuz:
This is to bring parity between 1809 and Windows Server 2016 and 2019. The reason behind this is that most modern apps can **not** be deployed to Windows Server 2016. By adding WinAppDriver to Windows 10 1809 Build, users will be able to deploy their modern applications to 1809 and use WinAppDriver to run UI automated tests within their CI pipeline. 

Request is similar to https://github.com/Microsoft/azure-pipelines-image-generation/pull/530.


DavidStaheli:
Thanks for this pull request, @hassanuz!  We've tried to keep the **Windows Container 1803** image as bare as possible - its primary use being **running containers** that already contain the tools people need.  But the **VS2017-Win2016** and **VS2019-Win2019** images are for adding common tools and running UI tests - and they already have WinAppDriver installed.  Do you need it on the **Windows Container 1803** image too?

PR #694, opened 2019-02-08
Updating az module for hosted agent in linux

20shivangi:


20shivangi:
@lokesh755 @dakale Can you please review PR? We are consuming this Az Module in AzurePowershellV4 task [Link ](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/AzurePowerShellV4/InitializeAz.ps1#L12).  This is corresponding user story[ Link ](https://mseng.visualstudio.com/AzureDevOps/_boards/board/t/Deployment/Stories).

dakale:
@DavidStaheli Can you help route this to the correct ecosystem

images/linux/scripts/installers/azpowershell.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Source the helpers for use with the script
+source $HELPER_SCRIPTS/document.sh
+
+# Install Azure CLI (instructions taken from https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
+sudo pwsh -Command 'Save-Module -Name Az -LiteralPath /usr/share/az_1.0.0 -RequiredVersion 1.0.0 -Force'
+
+# Run tests to determine that the software installed as expected
+echo "Testing to make sure that script performed as expected, and basic scenarios work"
+if ! pwsh -Command '$actualPSModulePath = $env:PSModulePath ; $env:PSModulePath = "/usr/share/az_1.0.0:" + $env:PSModulePath;
```
kmkumaran: Should we put in /usr/local/share? 
http://www.pathname.com/fhs/2.2/fhs-4.9.html

RoshanKumarMicrosoft: /usr/ usually contains files that come with the system distribution and /usr/local/ tree is free for the local administrator to manage.
/usr/share hierarchy is for all read-only architecture independent data file.
From my understanding in case of hosted image generation we are distributor this is the reason behind keeping inside /usr/share.
Even in case of miniconda installer we are keeping inside /usr/share/miniconda : [Link ](https://github.com/Microsoft/azure-pipelines-image-generation/blob/932baa06a9b9374459e4d176ece7f394cef9210c/images/linux/scripts/installers/miniconda.sh#L17) 
Also for Gradle/maven we are keeping inside /usr/share : [Link](https://github.com/Microsoft/azure-pipelines-image-generation/blob/d2605468761891a3371ac5bafafdf59a61fb9950/images/linux/scripts/installers/java-tools.sh#L55)
@dakale  Can you please suggest what is recommendation for hosted image generation ?

dakale: I dont have a strong opinion either way, so I would just favor consistency with existing patterns.

images/linux/ubuntu1604.json
```diff
@@ -139,6 +139,7 @@
             "type": "shell",
             "scripts":[
                 "{{template_dir}}/scripts/installers/android.sh",
+                "{{template_dir}}/scripts/installers/azpowershell.sh",
```
niadak: How we are enforcing this script to run after PowerShellCore is installed?

20shivangi: Resolved offline

RoshanKumarMicrosoft: They are managing the hierarchy they are running first set of [script](https://github.com/Microsoft/azure-pipelines-image-generation/blob/6c8667983465a7308ebb8755944f83e608c04249/images/linux/ubuntu1604.json#L92)  then they are running secondary or dependent [ script](https://github.com/Microsoft/azure-pipelines-image-generation/blob/6c8667983465a7308ebb8755944f83e608c04249/images/linux/ubuntu1604.json#L140). So if any thing depends on first set of script it should go in secondary script. Similar case of java and android.

PR #695, opened 2019-02-08
Added vcpkg root env variable to system variables in VS2017 and ubuntu images

vmapetr:
This PR implements `VCPKG_ROOT` system environment variable for win/linux images as was suggested in https://github.com/Microsoft/azure-pipelines-image-generation/pull/682#issuecomment-461494406 for https://github.com/Microsoft/azure-pipelines-image-generation/pull/682 C/C++: Add vcpkg 0.0.113 (or latest) (Linux, Windows).


sgrebnov:
:+1: I think this also makes sense for buuilding CMake projects where they need to know where `vcpkg.cmake` is located.

`cmake ../my/project -DCMAKE_TOOLCHAIN_FILE=${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake`

PR #698, opened 2019-02-11
Always install the latest release of Gradle on Ubuntu

### LINKED ISSUE #697: Update Gradle to 5.2.1 (or latest) on Ubuntu
Update Gradle to 5.2.1 (or latest) on Ubuntu

DavidStaheli:
This fixes #697.
This gets the latest released version number for Gradle and installs that version.


brcrista:
Did you run the canary builds with this change?

joshmgross:
Do we always want the latest version? What if there are breaking changes when switching major versions (i.e. Gradle 5 to Gradle 6)?

DavidStaheli:
Yes, tested with canary build.  It's okay that we always install the latest version (like we do on the Windows images) because any build that depends on a specific version uses the Gradle wrapper to download and build with that version.

images/linux/scripts/installers/java-tools.sh
```diff
@@ -37,11 +37,16 @@ ln -s /usr/share/apache-maven-3.6.0/bin/mvn /usr/bin/mvn
...
 
 # Install Gradle
-gradle_version="5.1.1"
-curl -sL https://services.gradle.org/distributions/gradle-"${gradle_version}"-bin.zip -o gradle-"${gradle_version}".zip
-unzip -d /usr/share gradle-"${gradle_version}".zip
-ln -s /usr/share/gradle-"${gradle_version}"/bin/gradle /usr/bin/gradle
-rm gradle-"${gradle_version}".zip
+wget -O gradleReleases.html https://gradle.org/releases/
+gradleUrl=$(grep -m 1 -o "https:\/\/services.gradle.org\/distributions\/gradle-.*-bin\.zip" gradleReleases.html | head -1)
+gradleVersion=$(echo $gradleUrl | sed -nre 's/^[^0-9]*(([0-9]+\.)*[0-9]+).*/\1/p')
```
lkillgore: I couldn't test this reg exp out, but does this work with just major + minor versions?  "5.2"?

(That is in the Gradle releases that way w/o a trailing '0')

DavidStaheli: Yes, here's what I tested:

```shell
root@12ccb20abf04:/agent# echo 5.2.1 | sed -nre 's/^[^0-9]*(([0-9]+\.)*[0-9]+).*/\1/p'
5.2.1
root@12ccb20abf04:/agent# echo 5.2 | sed -nre 's/^[^0-9]*(([0-9]+\.)*[0-9]+).*/\1/p'
5.2
root@12ccb20abf04:/agent# echo 5 | sed -nre 's/^[^0-9]*(([0-9]+\.)*[0-9]+).*/\1/p'
5
```

lkillgore:
I think this looks fine.  A nice-to-have would be to add some documentation on the expectations you have for the webpage that gets scraped.  (Reverse chronological order, specific download site, etc.)

PR #712, opened 2019-02-14
Spec arch to node.js in Vs2017-Server2016 image

pablonete:
Addresses #611 
Added node `arch`.
Updated node version to current one, see today's build: https://dev.azure.com/pablo-nunez/Runner/_build/results?buildId=304

### DIFF
--- a/images/win/Vs2017-Server2016-Readme.md
+++ b/images/win/Vs2017-Server2016-Readme.md
@@ -382,6 +382,7 @@ _Environment:_
 ## Node.js
 
 _Version:_ 10.14.2<br/>
+_Architecture:_ x64<br/>
 _Environment:_
 * PATH: contains location of node.exe<br/>
 
--- a/images/win/Vs2017-Server2016-Readme.md
+++ b/images/win/Vs2017-Server2016-Readme.md
@@ -381,7 +381,7 @@ _Environment:_
 
 ## Node.js
 
-_Version:_ 10.14.2<br/>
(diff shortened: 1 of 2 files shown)


pablonete:
@DavidStaheli @willsmythe May you guys review/merge this update? Thanks!

DavidStaheli:
@pablonete thank you for this clarification to the readme!  @alepauly @shadyysf @pasangh this change is accurate based on [Pablo's build](https://dev.azure.com/pablo-nunez/Runner/_build/results?buildId=304), but do you now have a special process for updating the readme, or can we merge this now?

alepauly:
> do you now have a special process for updating the readme, or can we merge this now?
We shouldn't update this by hand unless completely necessary. The image generation process generates this output automatically but we have to copy it into the proper readme file when it's done so either 1) the image includes the older version and we shouldn't update the readme, or more likely 2) the readme didn't get updated. @shadyysf generated the last one I think, do you still have the corresponding InstalledSoftware that gets generated?

bryanmacfarlane:
We use to merge it when the image was released.  (in all rings)

bryanmacfarlane:
Agreed on manual edits being an anti-pattern.  We should update the gen code.

pablonete:
Thank you all. So I've updated the script that generates the Readme, so the new Architecture is also auto-generated in the future. I've tested it locally. Can you guys take another look here @bryanmacfarlane @alepauly @DavidStaheli?

PR #713, opened 2019-02-14
Add the latest 4 versions of Boost 1.66.0 - 1.69.0 to Ubuntu and Windows images

max-lobanov:
Hello everyone,

This PR contains changes to preinstall 4 versions of boost (1.66.0, 1.67.0, 1.68.0, 1.69.0) both on Linux and Windows VM (VS2015 Server 2012 and VS2017 Server images with the following compilation options:
- debug and release
- x32 and x64,
- multithreaded
- link = static and shared

`bootstrap.bat` and `b2` should be run from the VS developer console to build boost on Windows images. We have implemented a few new methods to work with Visual Studio on image (get installation path, get toolset). All new functions were placed to the new helper file: `VSHelpers.ps1`

Currently compilation takes about **30 minutes for Linux** and **1h 20 minutes for Windows** for each boost version. So in total installation of 4 boost version significantly increases image generation time.
The size of built Boost is about 3GB for each Boost version

After additional investigation we have also noticed the following additional compilation options to expand the default configuration (not currently applied):
- variant: debug / release
- link: static / shared (AppVeyor has both versions pre-installed, so probably we want to to the same in order to provide smooth transition for potential customers)
- threading: single / multi
- address-model: 32 / 64


images/win/WindowsContainer1803-Azure.json
```diff
@@ -101,6 +101,12 @@
                 "{{ template_dir }}/scripts/Installers/Install-NodeLts.ps1"
             ]
         },
+        {
+            "type": "powershell",
+            "scripts":[
+                "{{ template_dir }}/scripts/Installers/Install-Boost.ps1"
+            ]
+        },
```
DavidStaheli: @max-lobanov I think the `WindowsContainer1803-Azure.json` image is the one Windows image on which we don't install most tools, including Boost.  Can you please confirm that @joebourneMS?  If so, we can undo edits to this file.

max-lobanov: Thank you for this point!
As I can see from the template - the most of the software is not installed on WindowsContainer image (including minGW and Visual Studio). So I think we should not install Boost too.
cc: @joebourneMS

austin-jensen:
@max-lobanov I think that really to be able to say we support the latest versions of boost we should at the very least support debug and release as well as 32 bit and 64 bit, since these are a common set of configuration options that people will use to test against. 

Those options seem more important, but all of the additional configurations pointed out in the pr description seem important to have added to properly support these boost versions in the C++ ecosystem.

@azureDaveOps  @DavidStaheli  any thoughts on this?

images/win/WindowsContainer1803-Azure.json
max-lobanov: Fixed

max-lobanov:
> @max-lobanov I think that really to be able to say we support the latest versions of boost we should at the very least support debug and release as well as 32 bit and 64 bit, since these are a common set of configuration options that people will use to test against.
> 
> Those options seem more important, but all of the additional configurations pointed out in the pr description seem important to have added to properly support these boost versions in the C++ ecosystem.
> 
> @azureDaveOps @DavidStaheli any thoughts on this?

We have enabled the following options for build: debug / release, x32/x64, static / shared build, multithreading. I have updated PR description to reflect this info.

At the moment PR is ready for review and we are working final validation.

Currently compilation takes about **30 minutes for Linux** and `1h 20 minutes for Windows` for each boost version. So in total installation of 4 boost version significantly increases image generation time.
The size of built Boost is about 3GB for each Boost version.

We believe it doesn't make sense to build the same code during each image generation. We suggest the following improvement: we can prebuild each boost version with the different toolsets (gcc, msvc-14.0, msvc-14.1), place artifact to some azure storage or toolcache and then download/unpack it during image generation. In this case Boost installation will take 5-10 minutes - the time to download [3GB * count of boost versions] instead of 2 hours for Linux images and 4 hours for Windows images.

@DavidStaheli , Could you let us know if this improvement looks good to you and if we need to take care about it in the scope of this feature.

DavidStaheli:
@max-lobanov that's a great idea to prebuild each version and download/copy it during image generation.  New versions are released every ~4 months. @azureDaveOps @vijayma @joebourneMS, does that sound okay?  If so, we can expand the scope of this feature.

max-lobanov:
> @max-lobanov that's a great idea to prebuild each version and download/copy it during image generation. New versions are released every ~4 months. @azureDaveOps @vijayma @joebourneMS, does that sound okay? If so, we can expand the scope of this feature.

@DavidStaheli ,
Let me share a bit more technical details of our suggestion:

- We need to store 3 archives (artifacts) for each Boost version 
    - Linux image - gcc toolset 
    - VS2015 image - msvc-14.0 toolset
    - VS2017 image - msvc-14.1 toolset 
 - The approximate size of each archive is about 2-3GB. So the each version of Boost will take about ~8GB in the storage for all VM images.
- I think we should not use toolcache to store these artifacts because the toolcache is downloaded entirely to the machine but we need only some parts (that matches the current toolset). 
- We can use some Azure blob storage to store artifacts.

We can automate and simplify process of creating new artifacts (when new version of Boost is released):
- We will create build definition that will download boost, build it and upload to Azure storage (This logic has already implemented in this PR so we will just reuse source code)
- Build definition will have the following parameters:
    - Version of boost
    - Toolset
- When new version of Boost is released we will just need to trigger this BD with params and waiting for build. As soon as build is passed and artifact is uploaded to the storage - we will update VM image template to add new version of Boost.
- To make sure that prebuilt binaries are compatible with image - we will use the same pool to build Boost (we will build GCC toolset in Ubuntu pool, msvc-14.0 toolset in VS2015 pool and etc...)

This approach will allow us to add new versions of Boost very quickly in future .

Issue #263, opened 2018-10-03
C/C++: Add the latest 2 versions of Boost 1.68.0 - 1.69.0 (Linux, Windows)


vtbassmatt:
A customer is asking for precompiled Boost libraries (debug and release) on the hosted images: https://github.com/apache/arrow/pull/2674

I know that C++ compilation, especially highly-templated tools like Boost, is slow, so from that angle, this feels like a win. I'm not sure about Boost's versioning story, though, so I don't know how many customers we can satisfy with a single or limited range of versions.

@azureDaveOps I think your team owns C++, could you have someone investigate?

vijayma:
Let us include the latest release version of Boost libraries on the hosted VS 2017 and Linux image. @DavidStaheli, @azureDaveOps , please have the imaging team work on a script for this.

DavidStaheli:
Downloads for different versions:
https://www.boost.org/users/history/

1.69.0 installation for Linux:
https://www.boost.org/doc/libs/1_69_0/more/getting_started/unix-variants.html

1.69.0 installation for Windows:
https://www.boost.org/doc/libs/1_69_0/more/getting_started/windows.html

PR #718, opened 2019-02-18
Fix Windows Server 2012 image generation issues

al-cheb:
1. Fix and improve Install-Go on Windows Server 2012 R2 VS 15:
  - **Issue:** Image generation for Windows Server 2012 fails due to missed script `Install-Go`
  - Improve Install-Go script to use the single script thought all windows images to avoid code duplications since script should be identical for all pools.
  - Define list of versions as template parameter and pass it to the script
      - "GO_VERSIONS=1.9.7, 1.10.8, 1.11.5",
      - "GO_DEFAULT=1.11.5"
2. **Issue:** AndroidSDK installation fails on Windows Server 2012 due to incorrect ndk folder.
**Fix:** add $ndk_root path for VS before 15.7.0 [Link](https://developercommunity.visualstudio.com/content/problem/248864/vs-1570-update-some-android-ndks-installed-into-ne.html)
3. **Issue:** Java installation fails due to missed tls param. 
**Fix:** Enable tls1.2 for Invoke-WebRequest
4. Fixed param typo in Install-Git.ps1: missing space between -package-parameters= "


images/win/vs2015-Server2012R2-Azure.json
```diff
@@ -212,6 +218,10 @@
         },
         {
             "type": "powershell",
+	    "environment_vars": [
+        	"GO_VERSIONS=1.9.7,1.10.8,1.11.5",
```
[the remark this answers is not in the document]
al-cheb: https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/vs2015-Server2012R2-Azure.json - should install Go, but Install-Go.ps1 missing from repo. 
216 -  "{{ template_dir }}/scripts/Installers/Install-Go.ps1"
338 - "{{ template_dir }}/scripts/Installers/Validate-Go.ps1"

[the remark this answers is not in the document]
sachinma: I just tried it on a public account and Go was not present in the path on Hosted Pool. It could be that the intention was to add Go support - but it never got added. We can in that case, choose to not install 1.9.7 and go ahead with the rest of the two versions. Thoughts?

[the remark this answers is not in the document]
max-lobanov: Hi @sachinma,
Go installation [was broken](https://github.com/Microsoft/azure-pipelines-image-generation/pull/691/files) for VS2015 pool recently. Script was moved but the template was not fixed properly so image generation process is broken for now. We are working on issue [adding Boost to the Linux / Windows images](https://github.com/Microsoft/azure-pipelines-image-generation/issues/263) so we fix image generation process.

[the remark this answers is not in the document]
sachinma: The fix looks good to me. Thanks for making the installation script generic! My question was that since Go is currently not present in the Hosted Image, probably we should skip 1.9.7 as it is already archived. We kept it in VS2017 image so as not to break the customers.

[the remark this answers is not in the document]
max-lobanov: Fixed!
Thank you for this point.

images/win/scripts/Installers/Install-JavaTools.ps1
```diff
@@ -11,6 +11,7 @@ $azulJDK11Uri = 'https://repos.azul.com/azure-only/zulu/packages/zulu-11/11.0.2/
 
 cd $env:TEMP
 
+[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
```
madhurig: Is this a request to use TLS1.2 or was it causing issues with Java tool setup?

images/win/scripts/Installers/Update-AndroidSDK.ps1
```diff
@@ -38,7 +38,12 @@ Expand-Archive -Path .\android-sdk-licenses.zip -DestinationPath 'C:\Program Fil
 $sdk_root = "C:\Program Files (x86)\Android\android-sdk"
 
 # The NDK is installed by Visual Studio at this location:
-$ndk_root = "C:\Microsoft\AndroidNDK64\"
+$ndk_root = "C:\Microsoft\AndroidNDK64"
+
+# For VS before 15.7.0
+if(-not (Test-Path $ndk_root)) {
+    $ndk_root = "C:\ProgramData\Microsoft\AndroidNDK64"
+}
```
madhurig: We don't install VS older than 15.7 on VS 2017. Why is this needed?

images/win/scripts/Installers/Install-JavaTools.ps1
max-lobanov: If we don't set it - it causes issue with `Invoke-WebRequest` PS command on Java tool setup on VS2015.
The reason why it causes issue on VS2015 but doesn't cause issue on VS2017 / VS2019 is that TLS is set by the other installer scripts that are run earlier during image generation.

madhurig: If we want to do it for VS2015, we should do it earlier in image generation. We should only focus on VS2017 and VS2019 for Windows, we are not planning to invest in VS2015 VM.

images/win/scripts/Installers/Update-AndroidSDK.ps1
max-lobanov: Yes, It is needed only for VS 2015 Server 2012 image since it causes build issues on it.

Per discussion with @DavidStaheli today, we should not invest more time in VS2015 since it is not planning to release. So does it make to merge these changes or should we just revert changes in this file (we can include only Install-Go changes to this PR)?

madhurig: Please revert these changes and include only the Go changes.

DavidStaheli: Thanks @max-lobanov and @al-cheb.  Apologies for not making this clearer from the start.

images/win/scripts/Installers/Install-JavaTools.ps1
max-lobanov: Reverted per discussion below

PR #719, opened 2019-02-18
Enable long path behavior on Windows images

### LINKED ISSUE #157: vs2017 packer should set the “Enable NTFS long paths policy”
I noticed when putting the agent in a fairly long path (c:\users\admin\downloads\vsts-agent-win-x64-2.134.2\...) that my build failed because of a long path produced in the _work folder.
I then moved it to another folder but also saw that the [NTFS Long Paths policy](https://www.ghacks.net/2016/05/27/microsoft-260-long-path-limit/) was not enabled. Not sure that would have helped, but for the future it would be good if the packer script would run a command to enable that flag so that we can break free from the terrible MAX_PATH problems of the 1990s.

DavidStaheli:
This fixes #157.

- [x] Tested on Windows Server 2016
- [x] Created canary test builds for VS2017 and VS2019

[Canary build for VS2017](https://dev.azure.com/buildcanary/CanaryBuilds/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=1646)
[Canary build for VS2019](https://dev.azure.com/buildcanary/CanaryBuilds/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=1647)

### DIFF
--- a/images/win/scripts/Installers/WindowsContainer1803/Initialize-VM.ps1
+++ b/images/win/scripts/Installers/WindowsContainer1803/Initialize-VM.ps1
@@ -4,7 +4,6 @@
 ##  Desc:  VM initialization script, machine level configuration
 ################################################################################
 
-
 function Disable-UserAccessControl {
     Set-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System" -Name "ConsentPromptBehaviorAdmin" -Value 00000000 -Force
     Write-Host "User Access Control (UAC) has been disabled."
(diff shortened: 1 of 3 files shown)


images/win/scripts/Installers/WindowsContainer1803/Initialize-VM.ps1
```diff
@@ -47,6 +45,9 @@ Write-Host "Setting local execution policy"
 Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope MachinePolicy -ErrorAction Continue | Out-Null
 Get-ExecutionPolicy -List
 
+Write-Host "Enable long path behavior"
+# See https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation
```
madhurig: We are enabling short name support before installing VS to support Xamarin AOT in VS2019 and VS2017: https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Vs2019/Install-VS2019.ps1#L21. Will there be a conflict between the 2 settings?

DavidStaheli: I can't produce a conflict, nor do the [docs ](https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#short-vs-long-names )mention potential conflict.

PR #725, opened 2019-02-19
Add Azure DevOps CLI extension install commands

gauravsaralMs:


images/linux/scripts/installers/azure-cli.sh
```diff
@@ -23,5 +23,7 @@ if ! command -v az; then
     exit 1
 fi
 
+az extension add -n azure-devops
```
hiyadav: Its good to have a step in validation script, that validates that the extension has been installed.

DavidStaheli:
This fixes #645.

gauravsaralMs:
I have made the changes, please review
I am trying to test the changes locally by following steps shared [here](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/docs/vs2017.md#create-a-vm-based-on-the-template-created-by-packer)

but on running `packer.exe build -var "client_id=$($spClientId)" -var "client_secret=$($spClientSecret)" -var "subscription_id=$($subscriptionId)" -var "tenant_id=$($tenantId)" -var "object_id=$($spObjectId)" -var "location=$($location)" -var "resource_group=$($rgName)" -var "storage_account=$($storageAccountName)" vs2017-Server2016-Azure.json`

I am getting error:
`Error initializing core: 1 error(s) occurred:

* required variable not set: install_password`

I can see that this variable was set because output is showing up for 
`Write-Host $install_password`

can someone help with this?

images/linux/scripts/installers/azure-cli.sh
gauravsaralMs: done

gauravsaralMs:
can someone point me instructions to test the changes for ubuntu?
I have tested it in standalone fashion, but would like to test complete thing end to end

hiyadav:
you can test it in similar fashion as windows. By running packer build command and passing values for these variables 
image

images/win/scripts/Installers/Validate-AzureDevOpsCli.ps1
```diff
@@ -0,0 +1,20 @@
...
+{
+    Write-Error "Azure DevOps Cli extension not present"
+    exit 1
+}
+else
+{
+    Write-Host "Azure DevOps Cli extension is present"
+}
+
+Add-SoftwareDetailsToMarkdown -SoftwareName 'Azure DevOps Cli extension' -DescriptionMarkdown $Description
```
hiyadav: Should we only document the version of azure devOps extension that is installed and not the usage information. Just like in linux.

gauravsaralMs: yeah there is a bit of parity issue, but I am trying to follow what azure cli has done
they also document just version in linux and a bit description in windows

RoopeshNair:
@sachina can you please review this?

images/win/scripts/Installers/Validate-AzureDevOpsCli.ps1
hiyadav: Yes, but that information is also about that path of installation. Whereas `az devops -h" contains the commands inside az devops, which is more of how the tool should be used. We can point to a documentation of usage if we must.

gauravsaralMs:
Sure, will make the required changes

images/win/scripts/Installers/Validate-AzureDevOpsCli.ps1
gauravsaralMs: makes sense, will make required changes

gauravsaralMs: Changed 
cc: @hiyadav

PR #728, opened 2019-02-19
Add Linux.CMake to the VS2017 image

mariusvniekerk:
Ensure that cmake projects can be built correctly with vs2017.  

This is related to https://github.com/conda-forge/conda-forge.github.io/issues/703 and occurs due to not having the cmake for linux box checked.


mariusvniekerk:
Alternatively we need VC14 in the new 2019 image

Issue #756, opened 2019-03-05
Python  3.7.2 (x64) is not installed


pasangh:
[packer-log882446267.zip](https://github.com/Microsoft/azure-pipelines-image-generation/files/2937251/packer-log882446267.zip)

**Describe the bug**
While running the packer build for latest VS2019 Image, I see a failure "Python 3.7.2 (x64) is not installed" 
![image](https://user-images.githubusercontent.com/32961811/53814604-e33cc680-3f2d-11e9-9c29-42f09f59852d.png)

**Actual behavior**
A description with steps to reproduce the issue.
1. Run packer for releases/VS2019/149.1

**Image impacted**
VS2019

pasangh:
Let me know if anymore info/logs are required here.

pasangh:
Fixed here - https://github.com/Microsoft/azure-pipelines-image-generation/pull/762

pasangh:
Fixed here - https://github.com/Microsoft/azure-pipelines-image-generation/pull/762

PR #762, opened 2019-03-06
Don't install python with VS, use versions in tools cache

madhurig:
https://github.com/Microsoft/azure-pipelines-image-generation/issues/756
- Remove Python 3 from VS installation (after discussing with Steve Dower)
- Use Python 3 from the tools cache, using python 3.6 for VS2017 and python 3.7 for VS 2019

Testing: Generated VS2019 and VS2017 successfully with these changes.

### DIFF
--- a/images/win/scripts/Installers/Validate-Python.ps1
+++ b/images/win/scripts/Installers/Validate-Python.ps1
@@ -15,9 +15,9 @@ else
     exit 1
 }
 
-$Python3Verison = $(python --version)
+$Python3Version = $(python --version)
 
-if ($Python3Verison -notlike "Python 3.*")
+if ($Python3Version -notlike "Python 3.*")
 {
     Write-Error "Python 3 is not in the PATH"
 }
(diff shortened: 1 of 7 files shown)


images/win/scripts/Installers/Validate-Python.ps1
```diff
@@ -11,12 +11,17 @@ if(Get-Command -Name 'python')
...
 else
 {
-     Write-Host "Python is not on path"
+    Write-Host "Python is not on path"
     exit 1
 }
 
 $Python3Verison = $(python --version)
 
+if ($Python3Verison -notlike "Python 3.*")
```
brcrista: could you fix this typo? ("verison")

images/win/scripts/Installers/Validate-Python.ps1
```diff
@@ -11,11 +11,16 @@ if(Get-Command -Name 'python')
...
 
-$Python3Verison = $(python --version)
+$Python3Version = $(python --version)
+
+if ($Python3Version -notlike "Python 3.*")
+{
+    Write-Error "Python 3 is not in the PATH"
+}
 
 $Python2Path = "C:\Python27amd64"
```
madhurig: This path exists, not sure what installs it, so I am leaving it as is.

Issue #734, opened 2019-02-23
windows-2019 VS2019 does not install all available windows SDK versions.


glennawatson:
**Describe the bug**
A number of the Windows SDKs arent' installed in the new hosted VS2019 installation.

**Expected behavior**
To have all the windows SDK installed that the 2019 installation supports:
Windows 10 Preview SDK (10.0.18323.0)
Windows 10 SDK (10.0.16299.0)
Windows 10 SDK (10.0.17134.0)
Windows 10 SDK (10.0.17763.0)

**Actual behavior**
Only has the default installation options here. Does not include the other Windows SDK options.

**Image impacted**
windows-2019

kaylangan:
The image now has Windows 10 SDKs:

- 16299
- 17134
- 17763
- 18362

Closing as complete.

PR #770, opened 2019-03-07
Add Windows SDK 16299 to VS2019 image

glennawatson:
This addresses 16299 missing from the Visual Studio 2019 image, as mentioned in https://github.com/Microsoft/azure-pipelines-image-generation/issues/734

### DIFF
--- a/images/win/scripts/Installers/Vs2019/Install-VS2019.ps1
+++ b/images/win/scripts/Installers/Vs2019/Install-VS2019.ps1
@@ -106,6 +106,7 @@ $WorkLoads = '--allWorkloads --includeRecommended ' + `
               '--add Microsoft.VisualStudio.Component.VC.TestAdapterForBoostTest ' + `
               '--add Microsoft.VisualStudio.Component.VC.TestAdapterForGoogleTest ' + `
               '--add Microsoft.VisualStudio.Component.VC.v141 ' + `
+              '--add Microsoft.VisualStudio.Component.Windows10SDK.16299 ' + `
               '--add Microsoft.VisualStudio.Component.Windows10SDK.17134 ' + `
               '--add Microsoft.VisualStudio.Component.Windows10SDK.17763 ' + `
               '--add Microsoft.VisualStudio.ComponentGroup.Azure.CloudServices ' + `


rmarinho:
@xamarin/xamarin.forms needs this also to build on VS2019 image

dansiegel:
This is a high impact issue would really appreciate if someone from the team could merge this!

DavidStaheli:
@vijayma, @azureDaveOps, and @alepauly can this be merged?

alepauly:
> This is a high impact issue would really appreciate if someone from the team could merge this!

@dansiegel - we're working on adding 18362 instead, that should be good for xamarin.forms, correct?

glennawatson:
A little confused isn't 18362 still in preview and the 2019 image using retail visual studio 2019?

I know there are plans to have a preview agent wouldn't it make more sense there. 

Also couldn't you have both to support backwards compatibility?

RLittlesII:
So VS 2019 is "stable" and 16299 is the latest stable version?  18362 is a preview version?  So why would we put the preview in the stable channel?  I don't think I mind having 18362 as an option, not the only option.  If I am not ready to take a dependency on preview parts, that puts me in a bad situation, no?

alepauly:
@glennawatson, @RLittlesII - you are both correct, apologies for some reason I thought 18362 had come out of preview. I'll merge. Thanks!

dansiegel:
@alepauly thanks for merging... keep in mind that adding support for 18362 which would be a good call as well doesn't really solve the root issue. Libraries need to be able to build based on older UWP SDK's because if I build for 18362 then any downstream app that consumes my library would only be able to target back to 18362. When a library builds for 16299 (which is where netstandard2.0 support was added) it allows app developers to target minimum support for that version and build for a newer version. Which of course is helpful on account of the fact that a lot of Enterprises are slow to adopt updates.

alepauly:
@dansiegel - agreed, but we have some disk size constraints that force us to be picky on these. I'm sure we'll end up adding 18362 but we'll probably need to make some other changes for that to happen. thanks.

glennawatson:
@alepauly how long does it take for these sorts of changes take to hit the agents?

dansiegel:
@alepauly if space is a concern I would keep the 16299 and swap out one of the other SDK's when you go to add 18362... 16299 is a critical point SDK where libraries are building against it because it was the point at which we got .netstandard2.0 support for UWP.

alepauly:
> @alepauly how long does it take for these sorts of changes take to hit the agents?

@glennawatson - We normally roll out new images every three weeks. These changes will make it in the images we start rolling out next week, so assuming no delays you'll see them in all Azure DevOps orgs during the week of May 27. It takes about a week from image generation for images to reach all orgs.

@dansiegel - thanks for the info, we'll keep that in mind when we're ready to add 18362.

Issue #769, opened 2019-03-07
VCPKG_ROOT should point to $HOME/.vcpkg


DavidStaheli:
This is an issue brought up by #682:

To fix the issue we need to rename VCPKG_ROOT variable to something like VCPKG_INSTALLATION_ROOT. So VCPKG_INSTALLATION_ROOT will point to the vcpkg installation dir and VCPKG_ROOT will point to the $HOME/.vcpkg by default.

As a temporary workaround customers need to override VCPKG_ROOT to $HOME/.vcpkg in their scripts before using vcpkg.

max-lobanov:
I am closing the issue since PR with fix was merged

taketwo:
Is this fix supposed to be available on hosted Azure agents already? If not, any estimates of when it will be?

At the moment I don't see any vcpkg environment variables at all. Here is my config that simply prints all variables:

```yaml
jobs:
  - job: vs2017
    displayName: Windows VS2017 Build
    timeoutInMinutes: 0
    pool:
      vmImage: 'vs2017-win2016'
    steps:
      - script: set
displayName: 'Print environment variables'
```

The output can be seen here: https://dev.azure.com/u7p11/pcl/_build/results?buildId=322.

DavidStaheli:
Hi @taketwo.  We closed this issue because the work is complete, but not available until the next image deployment.  You can click **Watch** to be notified of releases or view them here: https://github.com/Microsoft/azure-pipelines-image-generation/releases.

New releases happen about every 3 weeks right now, but we're working on releasing more often.

taketwo:
Thanks for the explanation!

PR #781, opened 2019-03-12
Rename VCPKG env variable to not override default VCPKG_ROOT

al-cheb:
Fix for the issue [#769](https://github.com/Microsoft/azure-pipelines-image-generation/issues/769) : Rename env `VCPKG_ROOT` to `VCPKG_INSTALLATION_ROOT` to prevent default environment variable overriding. `VCPKG_ROOT` is set by vcpkg and we should not override it so we need to use the different name for installation directory

 **Ubuntu 16.04 (vcpkg.sh) changes:**
       - Rename `$VCPKG_ROOT` to `$VCPKG_INSTALLATION_ROOT `
       - Add folder permissions ( `chmod 0777 -R $VCPKG_INSTALLATION_ROOT` ) to allow users to install vcpkg  packages in `$VCPKG_INSTALLATION_ROOT`

**Windows**(Install-Vcpkg/Validate-Vcpkg.ps1) changes: make the same changes (as for linux) to be consistent between images

### DIFF
--- a/images/linux/scripts/installers/vcpkg.sh
+++ b/images/linux/scripts/installers/vcpkg.sh
@@ -9,15 +9,15 @@
 source $HELPER_SCRIPTS/document.sh
 
 # Set env variable for vcpkg
-VCPKG_ROOT=/usr/local/share/vcpkg
-echo "VCPKG_ROOT=${VCPKG_ROOT}" | tee -a /etc/environment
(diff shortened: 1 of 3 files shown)


max-lobanov:
@madhurig , @azureDaveOps , Could you please review this PR?

PR #783, opened 2019-03-13
remove bad character introduced by copy paste for -

madhurig:
We are removing Python 3.* install to avoid conflicts with python installs in the tools cache in VS2019 and VS2017. There is a bad character in the script causing VS install to fail. Replacing that with the right hyphen character


pasangh:
@madhurig can you add a description? Thanks

PR #787, opened 2019-03-13
Update Ubuntu1604-README.md

faraonc:
Go 1.12 is now used according to https://github.com/Microsoft/azure-pipelines-image-generation/pull/779

### DIFF
--- a/images/linux/Ubuntu1604-README.md
+++ b/images/linux/Ubuntu1604-README.md
@@ -76,9 +76,7 @@ The following software is installed on machines in the Hosted Ubuntu 1604 pool
 - GNU C++ 7.4.0
 - Git (2.20.1)
 - Git Large File Storage (LFS) (2.6.1)
-- Go 1.9 (go version go1.9.7 linux/amd64)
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
This should be generated and not require a PR ...

JackMorrissey:
I don't think this is being generated and hasn't been for a while. If you take a look on Azure Devops for "What's on this image" it is pretty out of date. I just tried building .NET CORE SDK 2.2 with this image and it worked, while the documentation says it shouldn't.

Sorry to comment on an old random PR, this was the closest thing I could find to report this issue.

Compare https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md to the images to see the difference.

azure-devops-pipline-ubuntu-1604-2019-04-0
azure-devops-ubuntu-1604-whats-on-image-2019-04-05

kaylangan:
Thanks for the contribution. Our README's that list the installed tools are generated when we create the image and then merged into master. The difference you are seeing in the UI may be due to timing in rolling out the updates.

PR #805, opened 2019-03-25
Added built-in tests for toolcache

denis-panov:
**Changes**:
- Added Python and Ruby built-in test for toolcache.
- Changed testing logic: Now validation logic doesn’t use predefined (hardcoded) list of versions. It iterates through all existing versions in tool-cache folder and check them. Also we have checked that list of versions is checked in build canary tests.

### DIFF
--- a/images/linux/scripts/installers/test-toolcache.sh
+++ b/images/linux/scripts/installers/test-toolcache.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+################################################################################
+##  File:  test-toolcache.sh
+##  Team:  CI-Platform
+##  Desc:  Test Python and Ruby versions in tools cache
+################################################################################
+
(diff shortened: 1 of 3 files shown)


images/linux/scripts/installers/test-toolcache.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Fail out if any tests fail
+set -e
+
+AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
+
+# Python test
+if [ -d "$AGENT_TOOLSDIRECTORY/Python" ]; then
+   cd $AGENT_TOOLSDIRECTORY/Python
+   python_dirs=($(ls -d */))
```
brcrista: I think we should also fail if this list is empty

denis-panov: @brcrista I added this checking. Thanks!

Issue #778, opened 2019-03-11
Install-VS2017 fails since last week


riezebosch:
```log
    vhd: Running
    vhd: Using default tag: latest
    vhd: latest: Pulling from microsoft/dotnet-framework
    vhd: 3889bb8d808b: Already exists
    vhd: b479a92ce115: Already exists
    vhd: 9b1f172384ea: Already exists
    vhd: 848426354374: Already exists
    vhd: 7a39188c8674: Already exists
    vhd: 4588784ea70c: Already exists
    vhd: Digest: sha256:3c3fe0db41aa3500539a9667e0f1c2f6c2cc411bb658502bef8629f01f78b70e
    vhd: Status: Downloaded newer image for microsoft/dotnet-framework:latest
==> vhd: Provisioning with Powershell...
==> vhd: Provisioning with powershell script: C:\agent1\_work\51\s\images\win/scripts/Installers/Vs2017/Install-VS2017.ps1
    vhd: Enable short name support on Windows needed for Xamarin Android AOT, defaults appear to have been changed in Azure VMs
    vhd: Downloading Bootstrapper ...
    vhd: Starting Install ...
    vhd: Non zero exit code returned by the installation process : 1.
```

svarcoe:
I'm also running into this error. I logged into the VM and dug into the VS installer logs. I found the following error but not sure if it's related 

```
2019-03-12T17:01:49 : Error : Error: At most one command parameter can be specified
    at CommandLineErrorHandler.emit (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:652189)
    at CommandLineErrorHandler.emitError (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:651755)
    at CommandLineParser.extractAndNormalizeCommand (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:273082)
    at CommandLineParser.parse (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:269021)
    at Object.<anonymous> (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:675902)
    at __webpack_require__ (C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\main\StubMain.js:4:128)
[...]

PR #806, opened 2019-03-25
install vcredist140 latest

riezebosch:
because there are multiple dependencies on different versions down the chain. 

First [PHP](https://chocolatey.org/packages/php) installs `vcredist140 ≥ 14.11.25325.0` as a dependency. Later [openssl.light](https://chocolatey.org/packages/openssl.light) requires `vcredist140 ≥ 14.12.25810` and apparently `choco` doesn't install latest when a minimum required version is specified. The latter now fails because `vcredist140` doesn't support upgrade, apparently.

By installing `vcredist140` first both requirements for packages are satisfied.

### DIFF
--- a/images/win/vs2017-Server2016-Azure.json
+++ b/images/win/vs2017-Server2016-Azure.json
@@ -131,6 +131,10 @@
         },
         {
             "type": "powershell",
+            "inline": ["choco install vcredist140 -y"]
+        },
+        {
+            "type": "powershell",
             "valid_exit_codes": [
                 0,
                 3010
@@ -270,7 +274,7 @@
                 "{{ template_dir }}/scripts/Installers/Install-Go.ps1"
             ]
         },
-         {
+        {
             "type": "powershell",
             "scripts":[
                 "{{ template_dir }}/scripts/Installers/Install-PHP.ps1"


riezebosch:
Fixes #778

ryanspletzer:
Eagerly awaiting this one. Thank you!

azureDaveOps:
We are taking a look at this today. Thanks!

joebourneMS:
@shadyysf Can you review and merge?

ryanspletzer:
Eagerly awaiting this one -- currently causes an error in our packer builds.

images/win/vs2017-Server2016-Azure.json
```diff
@@ -129,6 +129,10 @@
                 "{{ template_dir }}/scripts/Installers/Vs2017/Update-DockerImages.ps1"
             ]
         },
+        {
```
shadyysf: Is this needed for the VS2019 image as well. If so, we should make a similar change to https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/vs2019-Server2019-Azure.json

riezebosch: I don't know because I don't build & use that one. But my guess is you do?!

shadyysf: Ok, I can follow up on that. @pasangh for fyi, in case the VS2019 image build fails because of vcredist we might need to take the same change to the VS2019 image. I will also cherry-pick this change to the 150 release branch.

pasangh: Thanks for the heads up Shady. I'll test out VS2019 to check for related issue.

I reviewed the PR and it seems good. I'll go ahead and approve it. Although it would be great if there was a **description on what testing has been done for the change**? Simply a check box with a tick for `Successfully built win server 16 VS2017 would be great`

PR #824, opened 2019-04-03
Install latest ansible from PPA:ansible/ansible

chadhellyea:
In order to use ansible[azure] you need a newer version of ansible than 2.0 which ships with 16.04


ameyaagashe:
Any update when updated Ansible version would be available on Azure Release pipelines - Ubuntu agent 16.04?

PR #831, opened 2019-04-04
Updating Az module V1.6.0

niadak:
Updating Az module V1.6.0


shadyysf:
@pasangh , here's another change that we might need to get into the VS2019 image.

PR #832, opened 2019-04-04
An additional article w/ ADO Hosted macOS Mojave VM Image description

Kutkovsky:
An initial article w/ Mojave VM Image description based on 03/30 rev. 1.
We plan that this particular VM Image will be used for the first deployment.


images/macos/macos-10.14-Readme.md
```diff
@@ -0,0 +1,545 @@
+# Azure Pipelines hosted macOS image
+
+The following software is installed on machines in the Azure Pipelines **macOS-10.14** VM image ('Hosted macOS' pool).
```
sergey-akhalkov: In order to be consistent with `'Hosted macOS High Sierra' pool` we need to add OS name here too: `'Hosted macOS Mojave' pool`

Kutkovsky: I use an existing Pool name. That is the question to the Pool's owners then

Kutkovsky: image

PR #840, opened 2019-04-08
Add Confirm flag to apt.sh

BobSilent:
in order to prevent the Question
You are about to add the following PPA to your system:
  ...
Press [ENTER] to continue or ctrl-c to cancel adding it


DavidStaheli:
Thank you, @BobSilent!

PR #856, opened 2019-04-12
Update VS2017 readme after AzureRM revert fix.

alepauly:
Update to VS2017 readme after fixes.


alepauly:
Self-approving merging to continue with deployment.

PR #875, opened 2019-04-20
Go: major version shouldn't have patch number

ItalyPaleAle:
This makes it possible to specify something like `GOROOT: '/usr/local/go1.12` for Go 1.12.

It now behaves similarly to how 1.11, 1.10 and 1.9 do.


DavidStaheli:
@sachinma and @shashankbarsin, I hope you don't mind if I merge this without your review.  It fixes an error I introduced when adding Go 1.12.1.  @ItalyPaleAle thank you!

sachinma:
Thanks @DavidStaheli for fixing this. I also noticed this and I was planning to fix this.

PR #880, opened 2019-04-22
Update Install-Ruby.ps1 and Validate-Ruby.ps1 scripts

al-cheb:
Issue:
Old hard-coded Ruby 2.5.0 was deleted from the tool cache that cause an error in the Validate-Ruby.ps1 

Fix:
Update Install-Ruby.ps1 and Validate-Ruby.ps1 to use the  latest Ruby 2.5.x version from the tool cache.

### DIFF
--- a/images/win/scripts/Installers/Validate-Ruby.ps1
+++ b/images/win/scripts/Installers/Validate-Ruby.ps1
@@ -47,7 +47,7 @@ $SoftwareName = "Ruby (x64)"
 $Description = @"
 #### $rubyVersionOnPath
 _Environment:_
-* Location:$rubyBinOnPath
+* Location: $rubyBinOnPath
 * PATH: contains the location of ruby.exe version $rubyVersionOnPath
 "@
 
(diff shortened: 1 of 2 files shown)


images/win/scripts/Installers/Validate-Ruby.ps1
```diff
@@ -38,21 +38,28 @@ else
...
+$rubyVersionOnPath = (Get-Command -Name 'ruby').Path
 
-_Location:_ C:\hostedtoolcache\windows\Ruby\2.4.3\x64\bin
+# Markdown Output
+$rubyVersionOutput = Get-ChildItem -Path $toolcacheRubyPath -Directory | Foreach-Object {
+	$rubyBinPath = Join-Path $_.FullName 'x64\bin'
+	$rubyVersion = Get-RubyVersion -rubyRootPath $rubyBinPath
 
-#### $rubyVersionOnPath
+	if ($rubyVersionOnPath.Contains($rubyBinPath)) {
```
max-lobanov: Here you go through all ruby versions in tool-cache. I think we have already do [the same](https://github.com/Microsoft/azure-pipelines-image-generation/blob/54c56a773c0101ba4b012f901b2194683230a16c/images/win/scripts/Installers/Validate-ToolCache.ps1)

I think we need to validate ruby on path in this file and validate tool-cache ruby in Validate-ToolCache.ps1

PR #900, opened 2019-05-03
Add Rust 1.34.1 support for VS2017, VS2019, Ubuntu1604

vmapetr:
This PR is based on PR https://github.com/Microsoft/azure-pipelines-image-generation/pull/653 and addresses the issue https://github.com/Microsoft/azure-pipelines-image-generation/issues/495 Add Rust 1.34.0 (or latest) (Linux, Windows):
- Implemented script to install Rust on Windows
- Implemented script to install Rust on Linux

**Details**
This PR is also includes implementation of [rustup](https://github.com/rust-lang/rustup.rs). Rustup is toolchain multiplexer that allows to install The Rust Programming Language from the official release channels, enabling you to easily switch between stable, beta, and nightly compilers and keep them updated.

**How it was tested**
- Run scripts manually on local boxes (Windows 10, Ubuntu 16.04).
- Via Azure VM provisioning and then verifying rustc, cargo, rustup.
`rustc ./helloworld.rs`
`rustup install nightly`
`cargo build`
- Canary Tests is created in Linux, VS2017, VS2019 folders..


myagley:
Thank you for this!

johnterickson:
Great to see this - thanks @vmapetr !

max-lobanov:
@madhurig , @alepauly , Could you please take a look and approve this PR if it looks good to you?

asottile:
this PR is awesome! :tada: -- I'm seeing that `rustc` / `cargo` are present on the windows images but not on the linux ones:

https://asottile.visualstudio.com/asottile/_build/results?buildId=719

(don't mind the "fail" -- I wanted to poke around without wasting time running my testsuite -- the info is in the failed job)

alepauly:
> this PR is awesome! 🎉 -- I'm seeing that `rustc` / `cargo` are present on the windows images but not on the linux ones:

@asottile : the Ubuntu 16.04 image that has this change is still rolling out to all organizations in Azure DevOps. I just checked and it's expected to reach your org by Tuesday (6/4). 

Thanks for the feedback!

asottile:
awesome! -- also seeing some oddities on windows too now that you mention it:

https://asottile.visualstudio.com/asottile/_build/results?buildId=720

```
E           CalledProcessError: Command: ('C:\\Program Files\\Rust\\.cargo\\bin\\cargo.exe', 'install', '--bins', '--root', 'c:\\temp\\pytest-of-unknown\\pytest-0\\test_rust_hook0\\0\\.pre-commit\\repo5ksthe\\rustenv-default')
E           Return code: 1
E           Expected return code: 0
E           Output: (none)
E           Errors: 
E               error: no default toolchain configured
```

alepauly:
Can you please create an issue with enough details to repro? I'll make sure the appropriate team takes a look at it. Thanks!

asottile:
yep! will do -- which project would you like me to attach that to?

alepauly:
Please attach it to this repo (azure-pipelines-image-generation). thanks again.

asottile:
done -- thank you again!  https://github.com/microsoft/azure-pipelines-image-generation/issues/988

alexcrichton:
FWIW I've found that this change has broken all my existing Rust projects that were on CI for a few reasons. Most of the breakage can be worked around, but I figured it'd be good to bring up here:

* Existing projects would have installed rustup manually and then used that to install a toolchain. By default though, the existing toolchain takes precedent so a manually installed toolchain is no longer the default.
* I've found that there's permission errors on some operations, so `rustup update --no-self-update` is required when updating toolchains or installing a new toolchain.
* The default toolchain semi-quickly goes out of date, but most projects I think keep up with the stable/beta/nightly channels in general
* Existing projects which assumed rustup was in `$HOME/.cargo/bin` (the default installation directory) now broke because it's installed in a different directory (with `RUSTUP_HOME` configured)
* This wasn't uniformly applied across all images so projects which test on OSX or the VS2015 images still have to manually install rustup, and it makes the installation script more complicated than it was before since it's trying to multiplex an existing rustup vs not.

I think this is a cool change and I doubt that any breakage was intended, but I figured I'd post and say that this is at least causing quite a speed bump in all the Rust projects I've gone on azure, I think I've had to reconfigure basically all of them to ensure that rustup is installed correctly. I also realize it's probably too late to revert this (now that projects are switching over to using the default), but figured it'd be good to keep in mind for future updates!

johnterickson:
Great points @alexcrichton - I admit I didn't look at this too closely and was just excited to have setting up a Rust build be a few steps easier.  Sounds like there are some follow-ups needed and I'd like to hear your feedback.

> Existing projects would have installed rustup manually and then used that to install a toolchain. By default though, the existing toolchain takes precedent so a manually installed toolchain is no longer the default.

I'm not sure how to add rustup and an existing toolchain without breaking usage of rustup.  Should we file a bug on rustup/rustup-init here in that it's behavior should be the same whether or not there is an existing install? Open to ideas here. 🤔

>I've found that there's permission errors on some operations, so rustup update --no-self-update is required when updating toolchains or installing a new toolchain.

I think the permission errors are because of the non-default RUSTUP_HOME location - e.g. Windows locks down Program Files. Does this sound right?

>The default toolchain semi-quickly goes out of date, but most projects I think keep up with the stable/beta/nightly channels in general.

I didn't follow the implication here - is that it's out-of-date and so not really helpful anyway? Should we keep rustup, but remove the default install (and would that fix point 1)?

>Existing projects which assumed rustup was in $HOME/.cargo/bin (the default installation directory) now broke because it's installed in a different directory (with RUSTUP_HOME configured)

I wonder if there is a reason for this. If not, my vote would be to stick with the default.

> This wasn't uniformly applied across all images so projects which test on OSX or the VS2015 images still have to manually install rustup, and it makes the installation script more complicated than it was before since it's trying to multiplex an existing rustup vs not.

Agreed. Would it make sense to bring those in parity first or to address some of the above first?

Thanks for this feedback.  If I see future Rust related PRs, I'll shoot an FYI out your way CC @rylev

alexcrichton:
> I'm not sure how to add rustup and an existing toolchain without breaking usage of rustup. Should we file a bug on rustup/rustup-init here in that it's behavior should be the same whether or not there is an existing install? 

I agree here in the sense that there's probably not a way to add rustup to existing images without breaking existing workflows. I think ideally though the end goal I think would be best would be to not have Rust installed by default but have a default task that could be used to install Rust. For example there's a [Node.js installer task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/node-js?view=azure-devops) which looks like:

```yml
- task: NodeTool@0
  inputs:
    versionSpec: '12.x' 
```

and it'd be awesome to do something like

```yml
- task: Rustup@0
  inputs:
    defaultToolchain: stable-msvc
```

That I think would be super easy to drop into a config file and would handle all the cross platform bits, printing versions, etc. If Rust were preinstalled on the image it could also handle the case where it's already installed and the toolchain matches, skipping an update. For channel-based updates (like `nightly` which change each day) it'd still download a new version.

> I think the permission errors are because of the non-default RUSTUP_HOME location - e.g. Windows locks down Program Files. Does this sound right?

I think? I'm not too too familiar with the permissions setup, but an example failure is this one -- https://dev.azure.com/alexcrichton/cc-rs/_build/results?buildId=555 -- where it looks like it couldn't remove `C:\Program Files\Rust\.cargo\bin\rustup.exe` during a self-update

> I didn't follow the implication here - is that it's out-of-date and so not really helpful anyway? Should we keep rustup, but remove the default install (and would that fix point 1)?

Oh no it's not like it's not useful or anything, I think most existing projects on Travis and AppVeyor though typically by default test with 'stable' as a toolchain specifier which changes over time. (many have an explicitly pinned version as well, though). If projects start relying on the default install on Azure images though it may lag behind the `stable` channel itself. I think it'd be best to track "stable" as-is in azure images, but regenerating the image so often might be a non-starter!

> >     Existing projects which assumed rustup was in $HOME/.cargo/bin (the default installation directory) now broke because it's installed in a different directory (with RUSTUP_HOME configured)
>
> I wonder if there is a reason for this. If not, my vote would be to stick with the default.

Oh this is all arguably badly written CI code anyway. Example changes this required was:

* https://github.com/rust-lang-nursery/libm/commit/d14680279d7f5e041046afe15736545630d5b549#diff-af93d9b2f70def02f68a90023bd8f407L21
* https://github.com/rustwasm/wasm-bindgen/commit/e24c03182b540e1c99c5382f0532db096a7c5c37#diff-7ebe1745303b78a07133c1793290ec0eL7

We should probably have like a `cargo --print-bin-dir` or something like that in Cargo itself, I mostly wanted to just point out that it was a possible point of breakage.

> Agreed. Would it make sense to bring those in parity first or to address some of the above first?

I don't mind too much on the best way forward from here myself. I've got the strategy down now for fixing the breakage in projects I see (and feel like I'm familiar with the ways in which it is currently changing). I suspect that further changes may continue breakage, but I think that's ok perhaps? I'm curious on your thoughts of not installing by default but having an easy-to-drop-in task to install Rust as well though. That wouldn't break anyone existing but would provide a way forward for everyone to simplify their configuration still!

---

> Thanks for this feedback. If I see future Rust related PRs, I'll shoot an FYI out your way CC @rylev

Thanks for your response as well! I don't mind helping out where I can for Rust-related changes as well!

johnterickson:
@alexcrichton I created a bug to keep tracki fo this https://github.com/microsoft/azure-pipelines-image-generation/issues/1012

Issue #774, opened 2019-03-08
Azure CLI version is not displayed correctly in generated readme file for image (VS2017/VS2019)


alepauly:
When generating the VS2017 and VS2019 images, the Validate-AzureCli.ps1 script is supposed to generate the output used for the readme file that describes the image spec. The output should include the version number. However, I noticed that in the last VS2017 I generated, the version was not printed out, see: 
https://github.com/Microsoft/azure-pipelines-image-generation/blob/releases/VS2017/149.1/images/win/Vs2017-Server2016-Readme.md#azure-cli

However, I verified that in the generated image Azure-Cli version 2.0.59 was properly installed (and reported by az --version.

**Image impacted**
VS2017 and VS2019 since they share the same script.

PR #908, opened 2019-05-06
Changed azure cli version extraction

hiyadav:
Fixing #774

### DIFF
--- a/images/win/scripts/Installers/Validate-AzureCli.ps1
+++ b/images/win/scripts/Installers/Validate-AzureCli.ps1
@@ -14,17 +14,13 @@ else
     exit 1
 }
 
-
-if( $($(az --version) | out-string)  -match  'azure-cli \((?<version>.*)\)' )
-{
-   $azureCliVersion = $Matches.version
-}
+$azureCliVersion = az -v | findstr azure-cli
 
 # Adding description of the software to Markdown
 $SoftwareName = "Azure CLI"
 
 $Description = @"
-_Version:_ $azureCliVersion<br/>
+_Version:_ $azureCliVersion
 _Environment:_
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Validate-AzureCli.ps1
```diff
@@ -14,17 +14,13 @@ else
     exit 1
 }
 
-
-if( $($(az --version) | out-string)  -match  'azure-cli \((?<version>.*)\)' )
-{
-   $azureCliVersion = $Matches.version
-}
+$azureCliVersion = az -v | findstr azure-cli
```
sachinma: whats the new syntax of the version?

hiyadav: The new syntax doesn't have braces before the version,
thus it is show like: 
**_az-cli       2.0.63_**

sachinma: Should we just remove the braces from the previous regex in that case

hiyadav: The previous regex has not been working from long, whereas this one is producing right results. Hence, I preferred using this one over altering the existing one. The new one is being used in other validation helper files.

sachinma: Only thing is the extra az-cli in the version field; should we remove that? I think that is probably fine; but see if it is easy to remove that part

hiyadav: Okay, I'll remove the extra _azure-cli_

hiyadav: done

PR #911, opened 2019-05-06
Cherry pick Ruby fixes into VS2019 release branch

alepauly:
Hosted tool cache now contains Ruby 2.4.5 and 2.5.3, update paths in validation.

This fix was already put in master but releases/VS2019/151.1 was branched out before it went in.


alepauly:
Not needed any longer since later 151 branches were branched off master that already contained this fix.

PR #924, opened 2019-05-08
update docker pull for windows server core

reaperhulk:
When calling `docker pull microsoft/windowsservercore` docker will attempt to pull the `latest` tag. This tag does not exist against that dockerhub repo. Instead it should pull `ltsc2019`.

### DIFF
--- a/images/win/scripts/Installers/Vs2019/Update-DockerImages.ps1
+++ b/images/win/scripts/Installers/Vs2019/Update-DockerImages.ps1
@@ -19,7 +19,7 @@ function DockerPull {
     $results
 }
 
-DockerPull microsoft/windowsservercore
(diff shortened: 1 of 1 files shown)


brcrista:
Hm, did this ever work?  Anyway, looks like the right fix.  Thanks!

It looks like [microsoft/nanoserver](https://hub.docker.com/_/microsoft-windows-nanoserver) also does not publish :latest and needs to be specified.  I can update that myself.

alex:
Are you able to share any estimate on how long it'll be before this is deployed? We're expecting (hoping!) this will have a big impact on our CI time.

brcrista:
It's usually every ~3 weeks.  @lokesh755 any idea when we'll roll out the next image?

PR #934, opened 2019-05-13
Update VS2019 readme after 151.2 generation.

alepauly:
Update the contents of  the readme file for VS2019 to include changes made during the generation of the image for the 151.2 version.


alepauly:
This won't be needed since issues with containers require regenerating the images.

PR #939, opened 2019-05-14
Update tags for nanoserver and windowservercore container images in VS2017

alepauly:
Container images in previous version were not OS compatible, switching for tags that will work with Windows Server 2016.


alepauly:
Got external review from container images owner, force approving to get images going.

Issue #914, opened 2019-05-07
Ubuntu 16.04 readme incorrectly lists Python, Ruby tools in the "Cached container images" section


willsmythe:
**Describe the bug**
The [readme](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md) for the software installed on the Ubuntu 16.04 image  incorrectly lists Python, Ruby, pip, etc in the "Cached container images" section:

![image](https://user-images.githubusercontent.com/2503052/57299556-e0cc2980-70a2-11e9-8cf0-9baf78be2b8b.png)

**Expected behavior**
Only a list of cached container images should be listed in this section.

Fix may just involve changing the ordering of steps in the packer .json for this image.

**Actual behavior**
See description.

**Image impacted**
ubuntu

kaylangan:
Fixed in [release branch](https://github.com/microsoft/azure-pipelines-image-generation/blob/releases/ubuntu/151.2/images/linux/Ubuntu1604-README.md). Will be ported back to master once the release is ready.

PR #944, opened 2019-05-15
Fix display of container items and merge latest Ubuntu readme file.

alepauly:
Update the Ubuntu 16.04 readme file with the latest manifest currently rolling out to production and fix the scripts that output the cached container images to avoid listing other things under the Container Images heading.


kaylangan:
Fixes #914

Issue #873, opened 2019-04-19
Add Windows 10 SDK Update version 1903


davidstaheli:
Windows 10 SDK Update version 1903 can be installed with the 'Individual components' section of the Visual Studio 2019 installer.

The SDK includes XAML Islands v1, whereas the version for Windows 10 version 1809 was in preview. There are also improvements to the Windows Subsystem for Linux commands, and Windows UI Library 2.1.

DavidStaheli:
@vijayma , @azureDaveOps, @kaylangan, @joebourneMS this is important to include in the next Windows 2019 release if possible.

jevansaks:
This would keep us from needing to install a custom SDK during our pipeline on the hosted agent and save 5 minutes on each build. Do you have an ETA for when this will happen?

amerkoleci:
Need this 18362 SDK as well here!

kaylangan:
Merged. Will release with M152.

jevansaks:
When will this be live? It looks like the agents in our instance still don't have this SDK installed.

shadyysf:
@jevansaks , could you please provide your organization name? I would like to investigate why the 152 version hasn't reached you yet.

jevansaks:
Sure, we’re running in the “ms” organization. Here’s an example pipeline: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=19260

shadyysf:
@jevansaks , ahhh I see what's happening. The M152 release of the VS2017 and VS2019 images has been skipped this sprint since we were working on automating the release of the images to make delivering spring payloads faster. The M153 release is rolling out at the moment and should have those changes. ETA for that is early next week. I apologize for the inconvenience this has caused.

jevansaks:
No problem, thanks for the update!

MichaelLHerman:
@kaylangan @shadyysf  Should the documentation on https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2019-Server2019-Readme.md have been updated to include 18362 as part of #953 , or does that only happen after 153 is rolled out to the public ring?   When is that scheduled to happen?   

Thanks

shadyysf:
@MichaelLHerman, we update the readme file in the release branch when the new image is generated. We only merge that back into master once the image has made it all the way through all of our production scale units. So the docs for the  M153 version for the VS2019 image are here 

[...]

PR #953, opened 2019-05-17
Add Windows 10 SDK 18362

kaylangan:
Addresses #873


kaylangan:
@alepauly 18362 is now officially released on the [semi-annual channel](https://docs.microsoft.com/en-us/windows/release-information/). if possible, I think we should pull this in before releasing M152 for VS2017/VS2019. What do you think?

PR #985, opened 2019-05-31
Added composer reference to download packages in parallel to speed up…

pimjansen:
Added this open plugin to speedup build timelines:
https://github.com/hirak/prestissimo

What actually does is run the downloads in parallel on where the actual install goes much faster

## Benchmark Example
Below times are for a `laravel/laravel` project
288s -> 26s

### DIFF
--- a/images/linux/scripts/installers/php.sh
+++ b/images/linux/scripts/installers/php.sh
@@ -244,6 +244,9 @@ php composer-setup.php
 sudo mv composer.phar /usr/bin/composer
 php -r "unlink('composer-setup.php');"
 
+# Composer parallel download
+composer global require hirak/prestissimo
+
 # Install phpunit (for PHP)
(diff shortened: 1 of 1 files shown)


alepauly:
@pimjansen Thanks for the contribution, seems like this would help a few builds but I rather not take a dependency for the base ubuntu1604 image on prestissimo since it could break at any time and it seems like it's easy enough for users with builds that have lots of package dependencies to set it up as part of their builds. Let me know if you think this is a must for the base image or if perhaps I'm misunderstanding something.

pimjansen:
@alepauly from that point i share your vision however it is also a huge benefit. The problem is that when installing it during runtime it will cost you about 30 seconds as well on which the benefit is much smaller. When having it installed native it will globally consume less which on MS side would also mean that the agent will be available earlier i assume

If I look at a general PHP project which are mostly build on frameworks we are talking about a lot of packages.

PR #993, opened 2019-06-03
Add temp_resource_group_name variable to templates

vmapetr:
To manage appropriate resource cleanup in azure-pipelines CI we need to implement `temp_resource_group_name` into image templates.

**Details:**
`temp_resource_group_name` - builder variable that contains name assigned to the temporary resource group that created during the build. **If this value not set by user, a random value will be assigned, so it will not affect other customers**.


shadyysf:
Approved.

Issue #990, opened 2019-06-01
Windows PHP Startup: Unable to load dynamic library 'curl' and 'mbstring'


andyli:
**Describe the bug**
Simply running `php --version` would print warnings about curl and mbstring extensions not found.

**Expected behavior**
No warning message.

**Actual behavior**

Example build log from [my build](https://dev.azure.com/onthewings/GitHub%20public%20projects/_build/results?buildId=49&view=logs&lineStart=12&jobId=92f38c5a-9166-5edb-492c-b1eb3738def5&taskId=ee7f0d65-fc2d-55a4-168c-59a46890f400):
```
##[error]PHP Warning:  PHP Startup: Unable to load dynamic library 'curl' (tried: C:\php\ext\curl (The specified module could not be found.), C:\php\ext\php_curl.dll (The specified module could not be found.)) in Unknown on line 0
##[error]PHP Warning:  PHP Startup: Unable to load dynamic library 'mbstri
##[error]n
##[error]g
##[error]'
##[error] 
##[error](
##[error]t
##[error]r
##[error]i
##[error]e
##[error]d
##[error]:
##[error] 
##[error]C
##[error]:
##[error]\
##[error]p
##[error]h
##[error]p
##[error]\
##[error]e
##[error]x
##[error]t
##[error]\
##[error]m
##[error]b
##[error]s
##[error]t
##[error]r
##[error]i
##[error]n
##[error]g
##[error] 
##[error](
##[error]T
##[error]h
##[error]e
##[error] 
##[error]s
##[error]p
##[error]e
##[error]c
##[error]i
##[error]f
##[error]i
##[error]e
##[error]d
##[error] 
##[error]m
##[error]o
##[error]d
##[error]u
##[error]l
##[error]e
##[error] 
##[error]c
##[error]o
##[error]u
##[error]l
##[error]d
##[error] 
##[error]n
##[error]o
##[error]t
##[error] 
##[error]b
##[error]e
##[error] 
##[error]f
##[error]o
##[error]u
##[error]n
##[error]d
##[error].
##[error])
##[error],
##[error] 
##[error]C
##[error]:
##[error]\
##[error]p
##[error]h
##[error]p
##[error]\
##[error]e
##[error]x
##[error]t
##[error]\
##[error]p
##[error]h
##[error]p
##[error]_
##[error]m
##[error]b
##[error]s
##[error]t
##[error]r
##[error]i
##[error]n
##[error]g
##[error].
[...]

PR #1014, opened 2019-06-11
Fix path to extensions in php.ini

alepauly:
Php install path was moved so the path for extension loading has to be changed in php.ini to avoid loading from the default (c:\php\ext). Instead, change php.ini to load from "ext" folder relative to executable.

Tested modified version of script in a build. Currently running full VS2017 and VS2019 images to verify build but feel confident that change is ok.


alepauly:
Addresses #990

PR #1021, opened 2019-06-17
Image generation should fail if docker pull fails

ds-ms:


images/win/scripts/Installers/Vs2017/Update-DockerImages.ps1
```diff
@@ -9,14 +9,12 @@ function DockerPull {
...
-    $j = Start-Job -ScriptBlock { docker pull $args[0] } -ArgumentList $image
-    while ( $j.JobStateInfo.state -ne "Completed" -And $j.JobStateInfo.state -ne "Failed" ) {
-      Write-Host $j.JobStateInfo.state
-      Start-Sleep 10
-    }
+    docker pull $image
 
-    $results = Receive-Job -Job $j
-    $results
+    if (!$?) {
```
bansalaseem: what does !$? mean?

bansalaseem:
Good fix

images/win/scripts/Installers/Vs2017/Update-DockerImages.ps1
ds-ms: $? is the last returned exit code. 
!$? -> non zero check

PR #1025, opened 2019-06-19
Add AWS, Google Cloud, and Zeit Now CLIs to Ubuntu 16.04

davidstaheli:


kaylangan:
We'll also want to add this to Ubuntu 18.04

PR #1054, opened 2019-07-03
fix extension install directory to a directory which is in homedrive …

gauravsaralMs:
fix extension install directory to a directory which is in homedrive and will be packed with image.

Can someone please help with detailed steps using which I can test these changes end to end?


kaylangan:
Will this break any existing usage?

// cc @alepauly

gauravsaralMs:
> Will this break any existing usage?
> 
> // cc @alepauly

not expected to break anything

PR #1055, opened 2019-07-03
Changes in installer script to pick up latest helm version

prebansa:
Change helm download url in azure-pipelines-image-generation.
Also version was hardcoded till now, changed it to pick up latest each time this script is run.
Workitem - https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1560636


prebansa:
@bansalaseem Please review

kaylangan:
@prebansa @bansalaseem Should we consider installing helm on Windows and Mac (@nrajpurkar ) too? Not saying in this PR, but in general.

prebansa:
@kaylangan we couldnt find such command for windows and osx which supports download of latest helm version.

PR #1060, opened 2019-07-03
Add popular container images to Ubuntu 16 and 18

pasangh:
List:

- Node:8, Node:10, Node:12
- Buildpack-deps:stretch
- Node:8-alpine, Node:10-alpine, Node:12-alpine
- Debian:8, Debian:9
- Debian:8-slim, Debian:9-slim
- Alpine:3.7, Alpine:3.8, Alpine:3.9, Alpine:3.10

### DIFF
--- a/images/linux/scripts/installers/docker.sh
+++ b/images/linux/scripts/installers/docker.sh
@@ -49,14 +49,6 @@ echo "Documenting Docker version"
 DOCKER_VERSION=`docker -v`
 DocumentInstalledItem "Docker ($DOCKER_VERSION)"
 
-echo "Installed docker images"
-docker images
-# Adding description of the software to Markdown
-
-$SoftwareName = "Docker images"
-$Description = @"
-The following container images have been cached:
-"@
-
-Add-SoftwareDetailsToMarkdown -SoftwareName $SoftwareName -DescriptionMarkdown $Description
-Add-ContentToMarkdown -Content $(docker images --digests --format "* {{.Repository}}:{{.Tag}} (Digest: {{.Digest}})")
+echo "Documenting Docker images installed"
+DOCKER_IMAGES=`docker images`
+DocumentInstalledItem "Docker ($DOCKER_IMAGES)"
--- a/images/linux/scripts/installers/docker.sh
+++ b/images/linux/scripts/installers/docker.sh
@@ -53,3 +53,12 @@ DocumentInstalledItem "Docker ($DOCKER_VERSION)"
 
 echo "Installed docker images"
 docker images
+# Adding description of the software to Markdown
+
+$SoftwareName = "Docker images"
+$Description = @"
+The following container images have been cached:
+"@
+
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/docker.sh
```diff
@@ -32,6 +32,19 @@ fi
...
+docker pull node:12-alpine
+docker pull debian:8
+docker pull debian:9
+docker pull alpine:3.7
+docker pull alpine:3.8
+docker pull alpine:3.9
+docker pull alpine:3.10
 
 ## Add version information to the metadata file
 echo "Documenting Docker version"
```
kaylangan: Do we need to change this to document the containers that are being cached?

pasangh: >We'll also want to add canary coverage.
Do you mean add testing to make sure these dockers are cached? If so, I'll add it.

> Do we need to change this to document the containers that are being cached?
Yeah, we need to add it. Thanks!

pasangh: Not sure how to make this a draft PR or mark it "Work in progress". I'll just update the description, thanks for the review though! :)

kaylangan: Yes, testing to make sure they are cached.

pasangh: Cool, I'll add it

images/linux/scripts/installers/docker.sh
```diff
@@ -32,8 +32,24 @@ fi
...
 DOCKER_VERSION=`docker -v`
 DocumentInstalledItem "Docker ($DOCKER_VERSION)"
+
+echo "Installed docker images"
+docker images
```
kaylangan: Do we need to use `DocumentInstalledItem` or `DocumentInstalledItemIndent` to get the right formatting?

images/linux/scripts/installers/docker.sh
```diff
@@ -32,8 +32,23 @@ fi
...
+docker pull alpine:3.10
 
 ## Add version information to the metadata file
 echo "Documenting Docker version"
 DOCKER_VERSION=`docker -v`
 DocumentInstalledItem "Docker ($DOCKER_VERSION)"
+
+echo "Documenting Docker images installed"
+DOCKER_IMAGES=`docker images`
+DocumentInstalledItem "Docker ($DOCKER_IMAGES)"
```
alepauly: do you want "DocumentInstalledItemIndent"?

ReneBosUnit4:
The description of this PR states that the slim version of both debian:8 and debian:9 will be cached, but only the non-slim versions are cached at this time. Is this on purpose or by accident?

Adding the slim versions of debian will improve pipeline performance for every PHP project because the default images as provided by PHP are based on the slim versions.

kaylangan:
@ReneBosUnit4 I believe we only cached the non-slim versions. If you could create a new issue to add the slim versions and an estimate of how much it improves build perf that'd be great!

PR #1064, opened 2019-07-04
Fix Path for Python3 from toolcache, remove Python2 VS installation

v-mitim:
**Changes:**
- Python3 Path didn't work because of the missing subfolder. Fixed by adding x64 to the path
`$python36path = $Env:AGENT_TOOLSDIRECTORY + '/Python/3.6*/x64'`
- If we want to use executables for windows toolcache installation then Python 3.4.* x86 have to be installed after x64 since x64 breaks installed x86 version. InstallTool function added to be able to run the separate installation of 3.4.4 x86 after the other versions installation/
- PYTHON_HOME env variable added to be consistent with [MMS provisioner factory script](https://dev.azure.com/mseng/AzureDevOps/_git/MMS.Provisioner?path=%2Fetc%2Ffactory%2Ffactory_UpdateWindowsPath.ps1&version=GBreleases%2Fm152)
- Python2 installation, which comes with VS, removed because it breaks the toolcache python 2.7.16 executable installer. Junction point added from C:\Python27amd64\ to the toolcache Python 2.
- Validate-Python script fixed. `Python3Version` check didn't work previously because of Python2 outputs the version to the stderr, so the test was passed in every case.

**How it was tested:**
- [VS2017 build](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9862928&view=results)
- [VS2019 build](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9859895&view=results)

### DIFF
--- a/images/win/scripts/Installers/Download-ToolCache.ps1
+++ b/images/win/scripts/Installers/Download-ToolCache.ps1
@@ -49,7 +49,9 @@ Get-ChildItem -Recurse -Depth 4 -Filter install_to_tools_cache.bat | ForEach-Obj
     }
 }
 
-InstallTool($Python34x86Path)
+if ($Python34x86Path) {
+    InstallTool($Python34x86Path)
+}
 
 Set-Location -Path $current
 
--- a/images/win/scripts/Installers/Download-ToolCache.ps1
+++ b/images/win/scripts/Installers/Download-ToolCache.ps1
@@ -4,6 +4,23 @@
 ##  Desc:  Download tool cache
 ################################################################################
 
+Function InstallTool
+{
+    Param
+    (
+        [System.Object]$ExecutablePath
+    )
+
+    Write-Host $ExecutablePath.DirectoryName
+    Set-Location -Path $ExecutablePath.DirectoryName
+    Get-Location | Write-Host
+    if (Test-Path 'tool.zip')
+    {
+        Expand-Archive 'tool.zip' -DestinationPath '.'
+    }
+    cmd.exe /c 'install_to_tools_cache.bat'
+}
+
 $SourceUrl = "https://vstsagenttools.blob.core.windows.net/tools"
 
 $Dest = "C:/"
@@ -21,16 +38,19 @@ $ToolsDirectory = $Dest + $Path
 $current = Get-Location
 Set-Location -Path $ToolsDirectory
 
+$Python34x86Path
 Get-ChildItem -Recurse -Depth 4 -Filter install_to_tools_cache.bat | ForEach-Object {
-    Write-Host $_.DirectoryName
-    Set-Location -Path $_.DirectoryName
-    Get-Location | Write-Host
-    if (Test-Path 'tool.zip')
-    {
-        Expand-Archive 'tool.zip' -DestinationPath '.'
+    #Python 3.4.* x86 have to be installed after x64 since x64 breaks installed x86 version
+    if ($_.DirectoryName -Match "Python\\3.4.*\\x86") {
+        $Python34x86Path = $_
     }
-    cmd.exe /c 'install_to_tools_cache.bat'
+    else {
+        InstallTool($_)
+    }
 }
+
+InstallTool($Python34x86Path)
+
 Set-Location -Path $current
 
 setx AGENT_TOOLSDIRECTORY $ToolsDirectory /M
--- a/images/win/scripts/Installers/Download-ToolCache.ps1
+++ b/images/win/scripts/Installers/Download-ToolCache.ps1
@@ -54,3 +54,7 @@ InstallTool($Python34x86Path)
 Set-Location -Path $current
 
 setx AGENT_TOOLSDIRECTORY $ToolsDirectory /M
+
+#junction point from the previous Python2 directory to the toolcache Python2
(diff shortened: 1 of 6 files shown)


images/win/scripts/Installers/Vs2017/Install-Python.ps1
```diff
@@ -26,4 +26,5 @@ if ($currentPath | Select-String -SimpleMatch $pythonDir.FullName)
 }
 
 Add-MachinePathItem -PathItem $pythonDir.FullName
-Add-MachinePathItem -PathItem "$($pythonDir.FullName)\Scripts"
\ No newline at end of file
+Add-MachinePathItem -PathItem "$($pythonDir.FullName)\Scripts"
+[System.Environment]::SetEnvironmentVariable('PYTHON_HOME', $pythonDir.FullName, [System.EnvironmentVariableTarget]::Machine)
```
brcrista: `PYTHON_HOME` doesn't actually do anything, I think this was just added by analogy to `JAVA_HOME`.  You can safely remove it.

images/win/scripts/Installers/Download-ToolCache.ps1
```diff
@@ -21,16 +38,23 @@ $ToolsDirectory = $Dest + $Path
...
+    if ($_.DirectoryName -Match "Python\\3.4.*\\x86") {
+        $Python34x86Path = $_
+    }
+    else {
+        InstallTool($_)
     }
-    cmd.exe /c 'install_to_tools_cache.bat'
 }
+
+InstallTool($Python34x86Path)
```
brcrista: check for non-null `$Python34x86Path`?  Today we will always have it, but this will make the script robust if we remove that version one day.

images/win/scripts/Installers/Vs2017/Install-Python.ps1
```diff
@@ -7,7 +7,7 @@
 
 Import-Module -Name ImageHelpers -Force
 
-$python36path = $Env:AGENT_TOOLSDIRECTORY + '/Python/3.6*'
+$python36path = $Env:AGENT_TOOLSDIRECTORY + '/Python/3.6*/x64'
```
brcrista: So if this was not working before, which version of Python was actually in PATH?

images/win/scripts/Installers/Download-ToolCache.ps1
```diff
@@ -21,16 +38,23 @@ $ToolsDirectory = $Dest + $Path
...
+
 Set-Location -Path $current
 
-setx AGENT_TOOLSDIRECTORY $ToolsDirectory /M
\ No newline at end of file
+setx AGENT_TOOLSDIRECTORY $ToolsDirectory /M
+
+#junction point from the previous Python2 directory to the toolcache Python2
+$python2Dir = (Get-Item -Path ($ToolsDirectory + '/Python/2.7*/x64')).FullName
+cmd.exe /c mklink /d "C:\Python27amd64" "$python2Dir"
```
brcrista: Since this is not covered in a canary, it would be good to test manually that this works correctly.

images/win/scripts/Installers/Vs2017/Install-Python.ps1
v-mitim: The same, but it was added during the deployment step with MMS provisioner script:
https://dev.azure.com/mseng/AzureDevOps/_git/MMS.Provisioner?path=%2Fetc%2Ffactory%2Ffactory_UpdateWindowsPath.ps1&version=GBreleases%2Fm152

images/win/scripts/Installers/Download-ToolCache.ps1
v-mitim: I've tested it on generated image, link works fine
image

images/win/scripts/Installers/Vs2017/Install-Python.ps1
brcrista: Oh I see, thanks.  In that case, we will need to leave that provisioner script in place until we roll out new images with this current change.  I'll leave a comment on that PR.

images/win/scripts/Installers/Download-ToolCache.ps1
brcrista: Thanks.  Can you just try executing something with `C:\Python27amd64\python` to make sure it is able to run and locate its libraries?  (I don't see any reason why it wouldn't work, I just want to cover our bases.)

images/win/scripts/Installers/Download-ToolCache.ps1
v-mitim: Thanks for good point! Fixed.

images/win/scripts/Installers/Vs2017/Install-Python.ps1
v-mitim: Thanks for the clarification. Removed.

images/win/scripts/Installers/Download-ToolCache.ps1
v-mitim: Sure!
Here it is - python 2.7.16 is definitely from the toolcache since VS python2 has 2.7.14 version
image

Issue #1023, opened 2019-06-19
devops extension for Azure Cli is not present in the images


atbagga:
**Describe the bug**
azure-devops extension for Azure Cli is not present in the images

**Expected behavior**
extension should be installed by default

**Actual behavior**
Devops extension is not installed by default

1. Create a cli/powershell task with inline script `az devops -h`
2. script should display devops help instead it fails.

**Image impacted**
ubuntu, window, mac

This was added in PR https://github.com/Microsoft/azure-pipelines-image-generation/pull/725 but it has not reflected in the images even till now.

atbagga:
Talked to @hiyadav  who suggested this could be due to different user accounts being used while running the install script and while using the agent later.

Azure CLI does install extensions to users' home directory by default so this could be a probable issue.

atbagga:
One of the suggested fix is to set the AZURE_CONFIG_DIR system wide environment variable to a common shared directory which all users have access to. 
This will end up sharing the extensions directory which is created on this path.

PR #1068, opened 2019-07-08
fix azure cli extension directory for ubuntu

gauravsaralMs:
Fixes : https://github.com/microsoft/azure-pipelines-image-generation/issues/1023

### DIFF
--- a/images/linux/scripts/installers/azure-cli.sh
+++ b/images/linux/scripts/installers/azure-cli.sh
@@ -27,6 +27,5 @@ fi
 DocumentInstalledItem "Azure CLI ($(az -v | head -n 1))"
 
 # Setup Azure Extension directory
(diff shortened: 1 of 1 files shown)


images/linux/scripts/installers/azure-cli.sh
```diff
@@ -25,3 +25,7 @@ fi
 
 # Document what was added to the image
 DocumentInstalledItem "Azure CLI ($(az -v | head -n 1))"
+
+# Setup Azure Extension directory
+echo "AZURE_EXTENSION_DIR=/usr/local/lib/azureExtensionDir" | tee -a /etc/environment
```
hiyadav: I hope the other extensions such as `az ml` also work with this ?

gauravsaralMs: yes, that will not change.

PR #1074, opened 2019-07-09
Remove fix for Python 3.4 installation on Windows

v-mitim:
Previously we've faced an issue - **Python 3.4.* x86** have to be installed after x64 since x64 breaks installed x86 version, but judging by the installation logs we don't need to change the installation order, x86 versions install after x64 by default. And besides, if we install Python 3.4 last we will get outdated py launcher that has limited functionality e.g. `py -0` doesn't work.

**Changes:**
- Separate installation of Python 3.4.4 x86 removed

### DIFF
--- a/images/win/scripts/Installers/Download-ToolCache.ps1
+++ b/images/win/scripts/Installers/Download-ToolCache.ps1
@@ -38,7 +38,6 @@ $ToolsDirectory = $Dest + $Path
 $current = Get-Location
 Set-Location -Path $ToolsDirectory
 
-$Python34x86Path
 Get-ChildItem -Recurse -Depth 4 -Filter install_to_tools_cache.bat | ForEach-Object {
     InstallTool($_)
 }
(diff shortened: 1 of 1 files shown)


images/win/scripts/Installers/Download-ToolCache.ps1
```diff
@@ -38,19 +38,8 @@ $ToolsDirectory = $Dest + $Path
 $current = Get-Location
 Set-Location -Path $ToolsDirectory
 
-$Python34x86Path
 Get-ChildItem -Recurse -Depth 4 -Filter install_to_tools_cache.bat | ForEach-Object {
```
brcrista: I think this is fine.  It would be good to leave a comment in the code pointing out that Python 3.4 x86 must be installed after x64, but this is just how it happens to work.

v-mitim: Thanks for good point, I've added the comment!

PR #1075, opened 2019-07-10
added latest az module version 2.3.2

20shivangi:


images/win/scripts/Installers/Install-AzureModules.ps1
```diff
@@ -245,7 +245,7 @@ else {
 # Download and unzip the stored AzurePSModules from the vstsagentools public blob
 $extractLocation = "C:\Modules"
 $azurePsUri = @(
-    "https://vstsagenttools.blob.core.windows.net/tools/azurepowershellmodules/AzPSModules.M151.20190418.18144.zip",
+    "https://vstsagenttools.blob.core.windows.net/tools/azurepowershellmodules/AzPSModules.M153.20190703.36963.zip",
```
kmkumaran: Can you also update the VS 2019 and VS 2017 read me with the Az module version details.

20shivangi: It is done.

kaylangan: In the future, can you only update the README if you're documenting a workaround for a bug with the currently deployed images? Right now, the README's are automatically generated when we generate a new image/release and then merged back into master when the deployment has reached all rings.

kaylangan: It looks like you replaced and existing version so now there are only a total of two versions saved/installed? Is that correct? Which versions? Might be helpful to add comments as to which zip is which version.

kmkumaran: @kaylangan - All the Az versions are available. The single zip has all the three Az modules. 

@20shivangi - you would want to address the feedback on adding comments.

Issue #931, opened 2019-05-10
Update/Add MFC support for ARM/ARM64


trevorlacey-msft:
**Tool infomation**
- Tool name: Optional VS2017 components for MFC on ARM64
- Add or update? Update
- Desired version: Any
- If this is an add request:
  - Brief description of tool: Builds applications with MFC dependencies for ARM64
  - URL for tool's homepage: None

**Areas and ecosystems**
- [x] C / C++

**Image impacted**
Windows VS2017

**Note**
I am currently unable to build applications with MFC dependencies on a hosted VS2017 image for ARM64.

`LINK : fatal error LNK1104: cannot open file 'atls.lib' [D:\a\1\s\my_project.vcxproj]`

I can successfully build on my own local machine after installing optional components:
- VC++ 2017 Libs for Spectre (ARM64)
- Visual C++ MFC for ARM64
- Visual C++ MFC support for ARM64 with Spectre Mitigations
- etc.

PR #1080, opened 2019-07-10
Adding ATL for ARM and ARM64.

PawelWMS:
Adding bits missing to build ARM project requiring `atls.lib` with both Spectre mitigations and without them. It seems that these bits are already part of VS2019 VM configuration and the VS2017 one contains the runtime components for ATL ARM/ARM64 as well.

This might also resolve issue #931.


kaylangan:
Thanks for the contribution!

PR #1086, opened 2019-07-15
Revert #1036 for nanoserver on Server 2016

adriaanthomas:
We found that `mcr.microsoft.com/windows/nanoserver:1803` is too recent for WS2016, where the Windows version (currently) is 10.0.14393.

There is no version tag mentioned for that Windows version in [this table](https://hub.docker.com/_/microsoft-windows-nanoserver#windows-images), but the previous label still works.

Without this, the image build breaks for VS2017/WS2016:
```
vhd: Installing mcr.microsoft.com/windows/nanoserver:1803 ...
vhd: 1803: Pulling from windows/nanoserver
vhd: Docker pull failed with a non-zero exit code
```


kaylangan:
@adriaanthomas thanks for the PR! A fix was recently merged in #1085

PR #1087, opened 2019-07-15
change ubuntu1604-readme download dir

yl2558:
The Ubuntu1604-README.md download destination was set to ./ instead of {{template_dir}}. When we run the Update Readme File step during the build image, it cannot find the modified readme file and push no changes in the created PR. 
Changed the download destination to fix this issue.


yl2558:
For ubuntu16, the metadata_file for the readme shouldn't be deleted after the readme is downloaded because the download may fail and it will be hard for us generate it manually.

PR #1102, opened 2019-07-23
Docker images for Ubuntu16 and Ubuntu18

pasangh:
Reference docker.sh in json files for Ubuntu16 and 18


images/linux/ubuntu1604.json
```diff
@@ -116,6 +116,7 @@
                 "{{template_dir}}/scripts/installers/build-essential.sh",
                 "{{template_dir}}/scripts/installers/clang.sh",
                 "{{template_dir}}/scripts/installers/cmake.sh",
+                "{{template_dir}}/scripts/installers/docker.sh",
```
alepauly: I think this needs to run afer docker-compose/moby?

PR #1120, opened 2019-07-26
Users/maoswald/fix ctype dot h

moswald:
This fixes the issue described in [devcommunity ticket 539158](https://developercommunity.visualstudio.com/content/problem/537209/missing-ctypeh-vc-includepath-on-windows-2019-agen.html). The WDK installer overwrites certain include paths and needs to come earlier in the order.

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -71,7 +71,9 @@ Function GenerateResourcesAndImage {
         [Parameter(Mandatory = $True)]
         [ImageType] $ImageType,
         [Parameter(Mandatory = $True)]
-        [string] $AzureLocation
+        [string] $AzureLocation,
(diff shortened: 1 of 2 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -108,7 +108,7 @@ Function GenerateResourcesAndImage {
     $spAppId = $sp.ApplicationId
     $spClientId = $sp.ApplicationId
     $spObjectId = $sp.Id
-    Sleep 60
+    Sleep 120
```
moswald: I kept having it fail on the next step, but doubling the sleep time fixed it for me. 🤷‍♂

alepauly: were you running locally? Maybe we should parameterize it, default to 30 secs? 2 minutes for no reason seems like a long time, I already felt like that with 1 min.

PR #1149, opened 2019-08-03
Improved GenerateResourceAndImage.ps1 PowerShell script

s-bauer:
When the resource group specified by the user is already existent it will be forcefully removed without user confirmation. This can cause serious trouble!
I added a confirmation prompt asking the user how to continue in that case. In addition I put in a Force flag that skips this new behavior and just forcefully removes the resource group.

Names of resource groups may contains special characters, that are not valid for storage account names. I removed the most common ones, so that the script does not fail in that case! Moreover I adjusted the script, so that it follows the naming conventions for azure provided by Microsoft.

### DIFF
--- a/helpers/GenerateResourcesAndImage.ps1
+++ b/helpers/GenerateResourcesAndImage.ps1
@@ -58,6 +58,9 @@ Function GenerateResourcesAndImage {
         .PARAMETER AzureLocation
             The location of the resources being created in Azure. For example "East US".
 
+        .PARAMETER Force
+            Delete the resource group if it exists without user confirmation.
+
         .EXAMPLE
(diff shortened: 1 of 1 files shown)


helpers/GenerateResourcesAndImage.ps1
```diff
@@ -110,10 +147,10 @@ Function GenerateResourcesAndImage {
     $spAppId = $sp.ApplicationId
     $spClientId = $sp.ApplicationId
     $spObjectId = $sp.Id
-    Sleep -Seconds $SecondsToWaitForServicePrincipalSetup
+    Start-Sleep -Seconds $SecondsToWaitForServicePrincipalSetup
```
alepauly: I didn't know about the difference between Start-Sleep and Sleep, good chance, thanks!

PR #1152, opened 2019-08-05
Tools cache for Ubuntu 1804

konradpabjan:
Ubuntu 1804 currently uses tools that are built on Ubuntu 1604. This has caused some incompatibilities and canaries to fail for Python.

In the tools cache, a new blob has been made for Ubuntu 1804 which is where Ruby and Python that are compiled on 1804 go.

Currently Ubuntu 1804 uses `https://vstsagenttools.blob.core.windows.net/tools/hostedtoolcache/linux`, which is the blob made by 1604.  With the following changes, 1804 will now use the new blob which is located at `https://vstsagenttools.blob.core.windows.net/tools/hostedtoolcache/ubuntu-1804`


images/linux/scripts/installers/1804/hosted-tool-cache.sh
```diff
@@ -0,0 +1,41 @@
...
+source $HELPER_SCRIPTS/document.sh
+
+# Fail out if any setups fail
+set -e
+
+# Download hosted tool cache
+AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
+echo "AGENT_TOOLSDIRECTORY=$AGENT_TOOLSDIRECTORY" | tee -a /etc/environment
+azcopy --recursive \
+       --source https://vstsagenttools.blob.core.windows.net/tools/hostedtoolcache/ubuntu-1804 \
```
konradpabjan: 1 line difference between 1604 and 1804. 

See new blob at: https://dev.azure.com/mseng/AzureDevOps/_git/CIPlat.Externals?path=%2Fhostedtoolcache%2Fubuntu-1804%2FPython&version=GBmaster Ruby and Python are the only 2 things added directly with this script.

konradpabjan:
All versions of Ruby are now in the new blob (2.3.7, 2.4.6, 2.5.5 and 2.6.2). For Python, versions 2.7.16, 3.5.7, 3.6.9 and 3.7.3 are in. 

We have decided NOT to support Python 3.4.10 on Ubuntu 1804 due to it reaching EOL and it requires OpenSSL 2.0 while Ubuntu 1804 comes with 2.1 by default.

kaylangan:
@konradpabjan why did we go with 3.7.3 instead of 3.7.4 for Python?

konradpabjan:
> @konradpabjan why did we go with 3.7.3 instead of 3.7.4 for Python?

3.7.4 is on its way. 3.7.3 was the default one that gets built by the tools cache when you run a new pipeline, I realized this after all the versions got in.

I'll be updating the blob hopefully tomorrow at the latest. I'm finishing a small improvement to the Python tools cache so I'll be rebuilding after my changes go in.

PR #1156, opened 2019-08-06
Update macos-10.14-Readme.md

al-cheb:
1. Update documentation about Xcode versions support
2. **Device Types** - this section was removed, as it duplicated other more detailed sections **iOS X.X**


images/macos/macos-10.14-Readme.md
```diff
@@ -106,48 +106,51 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
...
+| iOS 11.4                  | iphoneos11.4 | 9.4.1       |
+| iOS 12.0                  | iphoneos12.0 | 10.0        |
+| iOS 12.1                  | iphoneos12.1 | 10.1        |
+| iOS 12.2                  | iphoneos12.2 | 10.2(.1)    |
+| iOS 12.4                  | iphoneos12.4 | 10.3        |
+| iOS 13.0                  | iphoneos13.0 | 11.0        |
+| iOS Simulator 11.4        | iphonesimulator11.4 | 9.4.1    |
+| iOS Simulator 12.0        | iphonesimulator12.0 | 10.0     |
+| iOS Simulator 12.1        | iphonesimulator12.1 | 10.1     |
+| iOS Simulator 12.2        | iphonesimulator12.2 | 10.2(.1) |
```
Kutkovsky: Probably, it make sense to use the same way of Xcode version's pointing like you do on 216th string. List all available choices divided by comma.

images/macos/macos-10.14-Readme.md
```diff
@@ -190,55 +194,50 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
...
+| watchOS 5.0 | 16R5283q | 10.0       |
+| watchOS 5.1 | 16R591 | 10.1         |
+| watchOS 5.2 | 16T224 | 10.2, 10.2.1 |
+| watchOS 5.3 | 16U567 | 10.3         |
+| watchOS 6.0 | 17R5532d | 11.0       |
 
 #### Devices
 
-#### iOS 8.4
+#### How to install additional simulators?
```
Kutkovsky: This small help on simulators adding should be extended by our proposed way of using the particular simulator: a customer have to use the particular Xcode version (that contain a specific simulator inside) to build its app.
What do you think?

Kutkovsky: My suggestion is:

Each Xcode app has simulator inside its package. Instead of installing simulators you can just select the specific Xcode version. This is best practice.

`xcode-select-switch <path to your xcode app>`
This will set your default xcode to run the xcodebuild Then run the xcodebuild with your respective simulator.

`xcodebuild -configuration ${BUILD_TYPE} -target ${TARGET_NAME} -arch ${CPU_ARCHITECTURE} -sdk ${SIMULATOR_OR_IOS_SDK}`
Replace {SIMULATOR_OR_IOS_SDK} with the needed simulator value.

Or, if it is too large text, we can provide link to the MS article on how to switch Xcode version

images/macos/macos-10.14-Readme.md
```diff
@@ -89,7 +89,7 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
 
 | Version                | Build   | Path                            |
 |------------------------|---------|---------------------------------|
-| 11_beta                | 11M374r | /Applications/Xcode_11.app      |
+| 11.0                   | 11M374r | /Applications/Xcode_11.app      |
```
max-lobanov: Let's keep beta here not to confuse customers

images/macos/macos-10.14-Readme.md
```diff
@@ -190,55 +194,50 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
...
+| watchOS 5.1 | 16R591 | 10.1         |
+| watchOS 5.2 | 16T224 | 10.2, 10.2.1 |
+| watchOS 5.3 | 16U567 | 10.3         |
+| watchOS 6.0 | 17R5532d | 11.0       |
 
 #### Devices
 
-#### iOS 8.4
+#### How to install additional simulators?
+!!! - By default, we do not install all simulators for all Xcode versions due to compatibility issues that we have seen. To install a specific simulator for all Xcode versions, use `xcversion` tool:
```
max-lobanov: #### How to install additional simulators?
Image contains only default simulators for each Xcode version. To install a specific simulator for all Xcode version, use `xcversion` tool:
```
xcversion simulators --install='iOS 8.4'
```

images/macos/macos-10.14-Readme.md
```diff
@@ -402,7 +356,8 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
 - iPad Pro (11-inch)
 - iPad Pro (12.9-inch) (3rd generation)
 
-#### iOS 12.2, 12.4
+**iOS 12.2 [available for Xcode 10.2, 10.2.1]**  
+&nbsp;   &nbsp;   &nbsp;&nbsp;&nbsp;**12.4 [available for Xcode 10.3]**
```
sergey-akhalkov: Could you please clarify why we are using `&nbsp;` here and below?

images/macos/macos-10.14-Readme.md
```diff
@@ -106,139 +106,93 @@ Previously, Microsoft hosted Mac machines had JDKs pre-installed that were overl
...
 
-#### iOS 8.4
+#### How to install additional simulators?
+Image contains only default simulators for each Xcode version. To install a specific simulator for all Xcode version, use xcversion tool:
+```
+xcversion simulators --install='iOS 8.4'
+```
+**!** - The installation of some simulators may cause compatibility issues.
+
+#### iOS 8.4 [available for Xcode 9.4.1, 10.0, 10.1, 10.2, 10.2.1]
```
sergey-akhalkov: Do we really need this `[available for Xcode 9.4.1, 10.0, 10.1, 10.2, 10.2.1]` here and blow? It sound like we'll have to maintain this and the table above, I would suggest to remove this at all.

Issue #1166, opened 2019-08-12
Disable/NotInstall git credential manager


TingluoHuang:
**Describe the bug**
Git credential manager will prompt for credential and make the build looks like hanging...

**Expected behavior**
Disable or not install git credential manager

**Image impacted**
Windows

alepauly:
@kaylangan - thoughts on this? I think it makes sense to do it, I've heard of pain caused by this in the past.

kaylangan:
I have no objections to removing.

PR #1168, opened 2019-08-12
Disable Git Credential Manager on install

alepauly:
Some builds hang when Git auth fails and the GCM kicks in. Since the GCM is interactive, it will hang there but the user has no way to provide input. With this change, I'm disabling GCM on install to avoid this scenario.

Note that the choco package supports /NoCredentialManager but that only sets the variable at the user level. Since the user doing the install is different to that running any builds, we need to make sure we set it machine-wide.

### DIFF
--- a/images/win/scripts/Installers/Install-Git.ps1
+++ b/images/win/scripts/Installers/Install-Git.ps1
@@ -10,6 +10,9 @@ Import-Module -Name ImageHelpers
 # See https://chocolatey.org/packages/git
 choco install git -y --package-parameters="/GitAndUnixToolsOnPath /WindowsTerminal /NoShellIntegration"
 
+# Disable GCM machine-wide
+[Environment]::SetEnvironmentVariable("GCM_VALIDATE", "false", [System.EnvironmentVariableTarget]::Machine)
+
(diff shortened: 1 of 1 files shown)


alepauly:
Fixes #1166

images/win/scripts/Installers/Install-Git.ps1
```diff
@@ -10,6 +10,9 @@ Import-Module -Name ImageHelpers
 # See https://chocolatey.org/packages/git
 choco install git -y --package-parameters="/GitAndUnixToolsOnPath /WindowsTerminal /NoShellIntegration"
 
+# Disable GCM machine-wide
+[Environment]::SetEnvironmentVariable("GCM_VALIDATE", "false", [System.EnvironmentVariableTarget]::Machine)
```
TingluoHuang: what about `GCM_INTERACTIVE=false`?

better get reviewed by someone from VC client team.

alepauly: Found the [docs ](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Docs/Environment.md#gcm_interactive)and you are correct. Thanks!

PR #1196, opened 2019-08-24
Update ImageHelpers.psm1

sqlenders:
Exposed an extra function Install-VSIX to install Visual Studio Extensions

### DIFF
--- a/images/win/Vs2019-Server2019-Readme.md
+++ b/images/win/Vs2019-Server2019-Readme.md
@@ -111,6 +111,10 @@ The following workloads and components are installed with Visual Studio 2019:
 * Microsoft.VisualStudio.Workload.Universal
 * Microsoft.VisualStudio.Workload.VisualStudioExtension
 
+The following extensions are installed in Visual Studio 2019:
+* Microsoft Analysis Services Projects - version 2.8.17
+* Microsoft Reporting Services Projects - version 2.5.9
+
 ## WIX Tools
 
 _Toolset Version:_ 3.11.2318<br/>
--- a/images/win/scripts/ImageHelpers/ImageHelpers.psm1
+++ b/images/win/scripts/ImageHelpers/ImageHelpers.psm1
@@ -14,6 +14,7 @@ Export-ModuleMember -Function @(
     'Set-SystemVariable'
     'Install-MSI'
     'Install-EXE'
+    'Install-VSIX'    
     'Add-ContentToMarkdown'
     'Add-SoftwareDetailsToMarkdown'
 )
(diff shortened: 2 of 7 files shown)


sqlenders:
This change makes it possible to install extensions to Visual Studio 2019 very easily.
Basically the following lines are sufficent in a powershell script 

```
Import-Module -Name ImageHelpers -Force

$VSIXName = "Microsoft.DataTools.AnalysisServices.vsix"
$Url      = "https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ProBITools/vsextensions/MicrosoftAnalysisServicesModelingProjects/2.8.17/vspackage" 
$exitcode = Install-VSIX -Name $VSIXName  -Url $Url
```

In update #1195 I changed the ps1 file to make a command Install-VSIX.
In this update I added the rest from the files directly using GitHub Desktop.
In the file ImageHelpers.psm1 the command is exposed.
Then there are two installation scripts in /win/scripts/Installers/Vs2019/Extensions
I thought creating a new folder in Vs2019 is a good idea. 

The two installation scripts are for installing Analysis Services Projects and Reporting Services Projects extensions to Visual Studio 2019. This is not needed in older versions since we have BIDS or SSDT-BI.
I didn't install the Integration Services Project Extension because it is still preview. 

Finally the file vs2019-Server2019-Azure.json has been updated. I added the extensions just before the windows restart and after the last installation. 

Reviewers please pay attention to 
1) In images/win/scripts/ImageHelpers/InstallHelpers.ps1 in #1195. Function was copied from install-wix.ps1 and renamed to Install-VSIX to be consistent. Is this name correct ?
2) I think I also fixed a bug in the path with the single quote (') and the double quotes ("). When using single quotes the variables $ are not replaced. Is this correct ?
3) Since I copied it, I trust the error codes. Can you confirm these codes ? How can you test them ? 
4) Is /win/scripts/Installers/Vs2019/Extensions a good place for Extensions in Visual Studio 2019 ? 
5) About the files in /win/scripts/Installers/Vs2019/Extensions, are the filenames correct ? Is the description clear ? What should I use for team ? 
6) Is the location of the installation in /win/vs2019-Server2019-Azure.json correct ? Should it be earlier or later in the script ?
7) Should I write Validation Scripts too ? I can write a script that it checks for a VSIX file in C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions or lower. Should I check the version number too or is file exists enough ? 
8) Should I update the the /win/Vs2019-Server2019-Readme.md ? Add a section Visual Studio Extensions and then list the installed extensions with version numbers ? 
9) Should I add tags ? And which tags ? 
10) Should I check for duplicate installs ? And if detect them, generate a warning like 'Warning : duplicate installation detected. Installation skipped' or an error like 
'Error : duplicate installation detected. Installation aborted'. Or do nothing since we always start with a clean installation so there should not be any duplicated installations. 
11) How long will it take to review this ? Who will be reviewing this ? How does it work ? 

_This is my first change on a public github project. Please forgive me if I make mistakes. It is a bit overwhelming._

PR #1209, opened 2019-09-03
Updating Az PowerShell Module URL

20shivangi:


kaylangan:
Did the version change too? If so, from what to what? @20shivangi

20shivangi:
@kaylangan 
In this updated URL, I added one more version of Az module i.e. 2.6.0 along with the existing versions

kaylangan:
Got it, thanks! Can we make sure to add that info to the description and/or in the code comments in the future?

PR #1211, opened 2019-09-03
updatd script to install all dotnet sdks from releases-index.json

hiyadav:


images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -10,13 +10,32 @@ New-Item -Path C:\Temp -Force -ItemType Directory
...
+function GetSdkReleases()
+{
+    Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases-index.json' -UseBasicParsing -OutFile 'releases-index.json'
+    $dotnetChannels = Get-Content -Path 'releases-index.json' | ConvertFrom-Json
 
-#Filtering dotnet sdk and runtime versions which does not have "-" in their name, based on naming pattern they are either preview or rc versions
-$dotnetReleases = $dotnetReleases | Where-Object {!$_."version-sdk".Contains('-') } | Sort-Object {[Version] $_."version-sdk"}
+    # Consider all channels except preview channels.
+    # Sort the channels in ascending order
+    $dotnetChannels = $dotnetChannels.'releases-index' | Where-Object { !$_."support-phase".Equals('preview') } | Sort-Object { [Version] $_."channel-version" }
```
bishal-pdMSFT: Filter out eol channel as well

bishal-pdMSFT: Should we keep process of adding new channel manual? for example, even when 3.0 becomes non-preview, it should not get updated automatically as it can result into some unexpected issues for customers. rather we make an effort to add it manually and may be run some tests as well. this was the reason I did not do it for ubuntu. Thoughts?

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -10,13 +10,32 @@ New-Item -Path C:\Temp -Force -ItemType Directory
...
+
+    foreach ($dotnetChannel in $dotnetChannels)
+    {
+        $channelVersion = $dotnetChannel.'channel-version';
+        Invoke-WebRequest -Uri $dotnetChannel.'releases.json' -UseBasicParsing -OutFile "releases-$channelVersion.json"
+        $currentReleases = Get-Content -Path "releases-$channelVersion.json" | ConvertFrom-Json
+        # Consider all version except preview and rc. They contain '-' in their version number.
+        # Sort the versions in ascending order
+        $currentReleases = $currentReleases.'releases' | Where-Object { !$_."sdk"."version".Contains('-') } | Sort-Object { [Version] $_."sdk"."version" }
+        ForEach ($release in $currentReleases) {
```
bishal-pdMSFT: why not get all 'sdks' versions as well here only. That way you don't need to handle that separately

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,15 +72,50 @@ $dotnetReleases | ForEach-Object {
...
+            $sdks = $release.'sdks' | Where-Object { !$_.'version'.Contains('-') } | Sort-Object { [Version] $_.'version' }
+            ForEach ($sdk in $sdks)
+            {
+                InstallVersion -sdkVersion $sdk.'version'
+            }
+        }
+    }
+}
+
+function RunPostInstallaitonSteps()
```
bishal-pdMSFT: typo

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,15 +72,50 @@ $dotnetReleases | ForEach-Object {
...
-$cmdPath = "C:\Program Files\dotnet\userpath.bat"
-$cmdDotNetPath | Out-File -Encoding ascii -FilePath $cmdPath
+    $cmdPath = "C:\Program Files\dotnet\userpath.bat"
+    $cmdDotNetPath | Out-File -Encoding ascii -FilePath $cmdPath
+
+    # Update Run key to run a script at logon
+    Set-ItemProperty -Path "HKLM:\Software\Microsoft\Windows\CurrentVersion\Run" -Name "DOTNETUSERPATH" -Value $cmdPath
+}
+
+function BeginInstallation ()
```
bishal-pdMSFT: nit: no need to define this function as there is nothing after BeginInstallation(). You can just have three statements at top level itself.

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,15 +45,62 @@ $dotnetReleases | ForEach-Object {
...
 
-Add-MachinePathItem "C:\Program Files\dotnet"
-# Run script at startup for all users
-$cmdDotNetPath = @"
+function InstallAllValidSdks()
+{
+    Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases-index.json' -UseBasicParsing -OutFile 'releases-index.json'
+    $dotnetChannels = Get-Content -Path 'releases-index.json' | ConvertFrom-Json
+
+    # Consider all channels except preview channels.
```
bishal-pdMSFT: Fix the comment to say about eol as well

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,15 +45,62 @@ $dotnetReleases | ForEach-Object {
...
+
+    # Download installation script.
+    Invoke-WebRequest -Uri 'https://dot.net/v1/dotnet-install.ps1' -UseBasicParsing -OutFile 'dotnet-install.ps1'
+
+    ForEach ($dotnetChannel in $dotnetChannels)
+    {
+        $channelVersion = $dotnetChannel.'channel-version';
+        Invoke-WebRequest -Uri $dotnetChannel.'releases.json' -UseBasicParsing -OutFile "releases-$channelVersion.json"
+        $currentReleases = Get-Content -Path "releases-$channelVersion.json" | ConvertFrom-Json
+        $currentReleases = $currentReleases.'releases' | Where-Object { !$_.'release-version'.Contains('-') } | Sort-Object { [Version] $_.'release-version' }
```
bishal-pdMSFT: put comment to say that you are filtering '-' for preview and rc versions

images/win/scripts/Installers/Install-DotnetSDK.ps1
```diff
@@ -50,15 +45,62 @@ $dotnetReleases | ForEach-Object {
...
+        {
+            if ($release.'sdks'.Count -gt 0)
+            {
+                Write-Host 'Found sdks property in release: ' + $release.'release-version' + 'with sdks count: ' + $release.'sdks'.Count
+
+                # sort the sdks on version and remove preview/rc version from download list.
+                $sdks = $release.'sdks' | Where-Object { !$_.'version'.Contains('-') } | Sort-Object { [Version] $_.'version' }
+                ForEach ($sdk in $sdks)
+                {
+                    InstallSDKVersion -sdkVersion $sdk.'version'
```
bishal-pdMSFT: Do you first want to create a list of versions of 'sdk' and 'sdks' and then download them. There could be some non-unique versions i think.

images/win/vs2017-Server2016-Azure.json
```diff
@@ -151,6 +151,12 @@
             "type": "powershell",
             "inline": ["choco install vcredist140 -y"]
         },
+        {
+            "type": "powershell",
```
bishal-pdMSFT: is there a reason for change in this file?

hiyadav: VS installation also installs some versions of the .Net core sdk's. Which you used to cause problems when our script used to install versions older than them. 1.0 and 1.1 had a bug where they could not build cache when 2.0/2.1 was installed. 

We are no more installing 1.0 and 1.1 but this is to safe guard against any such issues that might arise.

PR #1212, opened 2019-09-03
Revert unifying linux/windows npm versions

brkerr:
Npm update was added to windows installers to unify the version installed with linux and windows. This overrode the already installed npm version packaged with nodelts.

Commit: https://github.com/microsoft/azure-pipelines-image-generation/commit/42a155dd6145a53416d79c45d9721f683e286187#diff-05e90063124f56972f7319f78d0b0db7


pablonete:
Linux is doing the same `npm i -g npm`, see [nodejs.sh](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/linux/scripts/installers/nodejs.sh#L16). Isn't that a potential source of failures as well?

PR #1216, opened 2019-09-05
Add Go 1.13 to Ubuntu 16.04, Ubuntu 18.04, VS2017, and VS2019

jonnylangefeld:
[go 1.13 was released](https://golang.org/doc/devel/release.html#go1.13) and it should be available in the agents

### DIFF
--- a/images/linux/scripts/installers/1604/go.sh
+++ b/images/linux/scripts/installers/1604/go.sh
@@ -33,5 +33,5 @@ function InstallGo () {
 InstallGo 1.9 1.9.7 false
 InstallGo 1.10 1.10.8 false
 InstallGo 1.11 1.11.12 false
-InstallGo 1.12 1.12.7 false
-InstallGo 1.13 1.13 true
+InstallGo 1.12 1.12.7 true
+InstallGo 1.13 1.13 false
--- a/images/linux/scripts/installers/1604/go.sh
+++ b/images/linux/scripts/installers/1604/go.sh
@@ -33,4 +33,5 @@ function InstallGo () {
 InstallGo 1.9 1.9.7 false
 InstallGo 1.10 1.10.8 false
 InstallGo 1.11 1.11.12 false
-InstallGo 1.12 1.12.7 true
+InstallGo 1.12 1.12.7 false
+InstallGo 1.13 1.13 true
--- a/images/linux/scripts/installers/1804/go.sh
+++ b/images/linux/scripts/installers/1804/go.sh
@@ -32,3 +32,4 @@ function InstallGo () {
 # Install Go versions
 InstallGo 1.11 1.11.12 false
 InstallGo 1.12 1.12.7 true
+InstallGo 1.13 1.13 false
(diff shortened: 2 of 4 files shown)


kaylangan:
Thanks for the contribution! Can you add this to Ubuntu 18.04, VS2017, and VS2019 too? Here's an example from when we added Go 1.12: https://github.com/microsoft/azure-pipelines-image-generation/pull/779

jonnylangefeld:
Thanks for the example @kaylangan, I added it for Ubuntu 18.04, VS2017, and VS2019

kaylangan:
Thanks, @jonnylangefeld! I'd like to get an approval from @davidstaheli or @alepauly and then I think we can merge this!

jonnylangefeld:
Awesome, thanks! Do you know when it will be released and when it will be available on the ADO agents?

images/linux/scripts/installers/1804/go.sh
```diff
@@ -31,4 +31,5 @@ function InstallGo () {
 
 # Install Go versions
 InstallGo 1.11 1.11.12 false
-InstallGo 1.12 1.12.7 true
+InstallGo 1.12 1.12.7 false
+InstallGo 1.13 true
```
bishal-pdMSFT: you need to specify 1.13 as second argument as well.

images/linux/scripts/installers/1604/go.sh
```diff
@@ -33,4 +33,5 @@ function InstallGo () {
 InstallGo 1.9 1.9.7 false
 InstallGo 1.10 1.10.8 false
 InstallGo 1.11 1.11.12 false
-InstallGo 1.12 1.12.7 true
+InstallGo 1.12 1.12.7 false
+InstallGo 1.13 1.13 true
```
bishal-pdMSFT: I am not too sure if 1.13 should be made default version. It has just released. We can give it some bake time and flip the default version later (may be next month)

images/linux/scripts/installers/1804/go.sh
jonnylangefeld: good catch, thanks!

images/linux/scripts/installers/1604/go.sh
jonnylangefeld: I reverted to 1.12.x as the default in all instances

kaylangan:
@bishal-pdMSFT I think this is ready when you have a chance to re-review. Can you create and self-assign an issue to update the default to 1.13 and make that change when you feel comfortable with it? Thanks!

bishal-pdMSFT:
> @bishal-pdMSFT I think this is ready when you have a chance to re-review. Can you create and self-assign an issue to update the default to 1.13 and make that change when you feel comfortable with it? Thanks!

Created [issue](https://github.com/microsoft/azure-pipelines-image-generation/issues/1219) to track this.

kaylangan:
@jonnylangefeld thanks again and congrats on your first contribution to this project! 🎉

jonnylangefeld:
no problem :)
do you know when this update will become available in the actual images used in azure dev ops? @kaylangan

kaylangan:
We will likely start rolling this out starting next week (at the latest).

// cc @alepauly

PR #1229, opened 2019-09-10
Added Kubernetes kind to windows and ubuntu1604 agents

Anumita:
kind is a tool for running local Kubernetes clusters using Docker container "nodes".

More on kind: https://github.com/kubernetes-sigs/kind

It will be useful for integration testing as we can spin up local kubernetes clusters


Anumita:
Canary test: https://github.com/microsoft/azure-pipelines-canary/pull/147

PR #1231, opened 2019-09-12
Decreasing disk size for Ubuntu images to support Ephemeral.

shadyysf:
This is needed to support running Ubuntu images on Ephemeral disks in Azure.

### DIFF
--- a/images/linux/ubuntu1604.json
+++ b/images/linux/ubuntu1604.json
@@ -43,7 +43,7 @@
             "image_publisher": "Canonical",
             "image_offer": "UbuntuServer",
             "image_sku": "16.04-LTS",
-            "os_disk_size_gb": "128"
+            "os_disk_size_gb": "86"
         }
     ],
     "provisioners": [
--- a/images/linux/ubuntu1804.json
+++ b/images/linux/ubuntu1804.json
@@ -43,7 +43,7 @@
             "image_publisher": "Canonical",
             "image_offer": "UbuntuServer",
             "image_sku": "18.04-LTS",
-            "os_disk_size_gb": "128"
+            "os_disk_size_gb": "86"
         }
     ],
     "provisioners": [


congyiwu:
I ran df -h on bbq-beets.  The amount of free space on the OS disk should drop from from ~73GB to ~31GB for Ubuntu16 and from ~79GB to ~37GB for Ubuntu18.

We can theoretically resize the disks back to 128GB in non-ephemeral pools (by setting OSDiskSizeGB), but we decide not to since we plan on doing ephemeral for all Ubuntu pools anyway

kaylangan:
Sounds like we can make this change given the numbers @congyiwu provided

shadyysf:
@alepauly, According to Congyi there's plenty of space left for both Ubuntu images. I'm waiting on a confirmation from @congyiwu after testing one of the actual images generated this way but we should be fine. The fact that I was able to generate the Ubuntu16 image without packer failing due to running out of disk space is also a good sign that we're good.

congyiwu:
@shadyysf the MMS.M0.Ubuntu16.0.vhd test image you sent me worked, so this change should be good to merge

shadyysf:
@congyiwu, feel free to approve the pr and I'll merge it.

congyiwu:
@shadyysf I approved it a few minutes after I made that comment 18 minutes ago once I realized I was a reviewer

congyiwu:
BTW, if someone needs to be unblocked ASAP, they can free up disk space by deleting files in the beginning of their build.  There's not a great existing list of where things are, but I was able to piece together some candidates w/ du -sh on Ubuntu16:

Also #1242 should free up 3.7GB by running `sudo apt-get clean` during image generation

```
# Save 3.7 GB /var/cache/apt (with no consequences unless you need to uninstall and reinstall a package):
sudo apt-get clean

# Save 8.3 GB if you don't need Android:
sudo rm -rf /usr/local/lib/android

# Save 1.7 GB if you don't need Boost C++ Libraries:
sudo rm -rf /usr/local/share/boost

# Save 1.3 GB if you don't need Go:
sudo rm -rf /usr/local/go*

# Save 14 GB if you don't need .NET Core:
sudo rm -rf /usr/share/dotnet

# Save 1.1 GB if you don't need Rust:
sudo rm -rf /usr/share/rust
```

Issue #1208, opened 2019-09-02
Ubuntu Az Module missing


dirkslab:
**Describe the bug**

According to Linux Image readme Az Module should be available.
Readme: https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md

The following software is installed on machines in the Hosted Ubuntu 1604 pool
Az Module (1.0.0)
Az Module (1.6.0)

**Expected behavior**

run az module command in powershell core, without needing to install az module.

**Actual behavior**

Name                           Value
PSVersion                      6.2.2
PSEdition                      Core
GitCommitId                    6.2.2
OS                             Linux 4.15.0-1055-azure #60-Ubuntu SMP Thu Aug …
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

Connect-AzAccount : The term 'Connect-AzAccount' is not recognized as the name of a cmdlet, function, script file, or operable program

workaround: 
  - pwsh: |
      Install-Module Az -Force -AllowClobber
    displayName: 'install az module'

**Image impacted**

ubuntu-latest and ubuntu-16.04

RoopeshNair:
@20shivangi / @niadak

niadak:
The Az Modules are saved in /usr/share/az_{version} folder. This is to support multiple Az module version side by side. You need to update the $env:PSModule path with the folder path to use the module.

Currently available versions are /usr/share/az_1.0.0, /usr/share/az_1.6.0, /usr/share/az_2.3.2 and /usr/share/az_2.6.0.

BTW, recommened way to run Azure PowerShell scripts in pipeline is using AzurePowerShell task.

dirkslab:
Thanks, will test this out. Will post response for others on the process
when done.

"BTW, recommend way to run Azure PowerShell scripts in pipeline is using
AzurePowerShell task." ...

 ...thanks, at the time I swapped from Task as it did not do what I needed
to do. Might revisit as updates to the tasks might now make them
sufficient.

dirkslab:
have attempted the following based on feedback above in yml pipeline template. All subsequent az calls still fail. Any ref to documentation would be much appreciated.

attempted:
$env:PSModulePath = $env:PSModulePath + ";/usr/share/az_2.6.0"
[...]

PR #1241, opened 2019-09-18
Users/shmittal/just check

20shivangi:


images/linux/Ubuntu1604-README.md
```diff
@@ -200,6 +200,8 @@ the
 - Android ConstraintLayour 1.0.1
 - Az Module (1.0.0)
 - Az Module (1.6.0)
+- Az Module (2.3.2)
```
niadak: Should we call out that this modules are not installed and is saved in the specific path. User needs to update the PSModulePath to use them?

niadak: There is a open issue for the same https://github.com/microsoft/azure-pipelines-image-generation/issues/1208

20shivangi: Done

20shivangi:
@kaylangan Can you tell me what I should do to update the README file ?

Issue #159, opened 2018-06-07
Run-Antivirus.ps1 sometimes results in 0x800106ba Windows Defender error


ryanspletzer:
This seems non-deterministic and will work sometimes, but is hard to test because it's at the very end of the build... which can take about 4 hours. If I move it to the front end of the build this step tends to work just fine.

When it gets to the [MsMpCmd.exe](https://github.com/Microsoft/vsts-image-generation/blob/master/images/win/scripts/Installers/Vs2017/Run-Antivirus.ps1#L11) step it may fail with:

```
CmdTool: Failed with hr = 0x800106BA. Check C:\Users\ADMINI~1\AppData\Local\Temp\MpCmdRun.log for more information
```

And after that fails [Set-MsMpPreference](https://github.com/Microsoft/vsts-image-generation/blob/master/images/win/scripts/Installers/Vs2017/Run-Antivirus.ps1#L15) also fails with:

```
+ Set-MpPreference -ScanAvgCPULoadFactor 5 `
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (MSFT_MpPreference:root\Microsoft\...FT_MpPreference) [SetMpPreference],
   CimException
    + FullyQualifiedErrorId : HRESULT 0x800106ba,Set-MpPreference
```

This is right after the machine comes back up from a reboot....

I'm hypothesizing if building in a pause step to allow the Windows Defender service to get "warmed up" after the reboot may help?

Have you guys ever seen this before when you run it?

lokesh755:
I've generated images with current repo many times, but I've never seen this issue. Did you change any builder settings ? How often you see this problem ? If it's too often can you add packer param "[-on-error=abort](https://www.packer.io/docs/commands/build.html)" , which will avoid packer to delete VM in case of any error . So that you can debug more easily and also collect logs.

lokesh755:
Are you able to collect the logs ?

ryanspletzer:
I just let it run again and it got to the antivirus scan part and made it in successfully this time with the code as-is and is in the process of scanning. Not sure what caused that to happen a few times... it may be non-deterministic and hard to reproduce, so we can probably close this out.

By the way, how long does the Run-Antivirus.ps1 step typically take for you guys? I see it take about 8 hours for me.

ryanspletzer:
[...]

PR #1259, opened 2019-09-25
Wait for windows defender to be running

alepauly:
When setting preferences on Windows Defender, if the windefend service is not running, it will fail with obscure errors. To resolve this I'm adding a wait until the windefend service is found running before proceeding to set the preferences and run a scan to finalize image generation.


domyd:
Looks like this might be a fix for #159.

PR #1265, opened 2019-09-26
Adding latest Az modules in hosted ubuntu

20shivangi:


images/linux/scripts/installers/1604/azpowershell.sh
```diff
@@ -23,11 +25,23 @@ if ! pwsh -Command '$actualPSModulePath = $env:PSModulePath ; $env:PSModulePath
     if (!(get-module -listavailable -name Az.accounts)) {
         Write-Host "Az Module was not installed"; $env:PSModulePath = $actualPSModulePath; exit 1
     }
+    $env:PSModulePath = $actualPSModulePath
+    $actualPSModulePath = $env:PSModulePath ; $env:PSModulePath = "/usr/share/az_2.3.2:" + $env:PSModulePath;
```
niadak: Should we consider to create a single file and use that for both image versions to avoid duplicate code?

20shivangi: Will take it in separate PR

PR #1273, opened 2019-09-27
Users/hiyada/dot net sdk installation with fix

hiyadav:


hiyadav:
@alepauly 
I ran the CI Build for these changes. Ran successfully finally: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10514356 

**The issue** seems to be that if the .Net Core version to be installed with VS2019 is already installed, then there is some abnormalities. I had earlier made changes to install .Net Core SDKs before VS2019 and just reverting that does the magic.

_Do you think it is a bug or something worth investigating in VS2019 installer ?_

alepauly:
> _Do you think it is a bug or something worth investigating in VS2019 installer ?_

@hiyadav if you have an easy minimal repro, the VS team can probably use the report. It's a very odd bug.

Thanks for the fix!

PR #1275, opened 2019-09-27
Update packer config to support private_virtual_network_with_public_ip

lupine86:
These config changes will allow for building packer images using the specified private virtual network with a public IP.

https://www.packer.io/docs/builders/azure.html#private_virtual_network_with_public_ip


images/linux/linux_preview.json
```diff
@@ -8,6 +8,10 @@
         "resource_group": "{{env `ARM_RESOURCE_GROUP`}}",
         "storage_account": "{{env `ARM_STORAGE_ACCOUNT`}}",
         "location": "{{env `ARM_RESOURCE_LOCATION`}}",
+        "virtual_network_name": "{{env `VNET_NAME`}}",
+        "virtual_network_resource_group_name": "{{env `VNET_RESOURCE_GROUP`}}",
+        "virtual_network_subnet_name": "{{env `VNET_SUBNET`}}",
+        "private_virtual_network_with_public_ip": "{{env `PRIVATE_VIRTUAL_NETWORK_WITH_PUBLIC_IP`}}",
```
alepauly: safe to assume that not having this value falls back to previous behavior, right?

lupine86: Yep -tested it and it fell back to previous behavior when the parameters are not specified.

Issue #270, opened 2018-10-10
C/C++: Add/update to LLVM 7.0.0 (or latest) x64 (Linux, Windows)


davidstaheli:
The VS2017 image needs LLVM and Clang

cstratopoulos:
Hey was about to open my own issue then saw this one. Having LLVM on Windows pre-installed would be great, but would it be possible to also add the [LLVM compiler toolchain extension](https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.llvm-toolchain#overview) ? This would make it a lot easier to use `clang-cl` with the [CMake task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/cmake?view=azure-devops), one hopes e.g., by being able to pass `-T "llvm"` to `cmakeArgs`

nicolasnoble:
Same with the vs2019 image actually.

Keats:
I've just ran into that as well, a project update now requires LLVM which is built-in in Appveyor but is not available in Azure Pipelines.
This would have been a hard blocker for me if that update happened a few weeks ago, when I was investigating Azure Pipelines.

reupen:
VS 2019 now comes with LLVM as an optional component. Is it possible for the `Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Llvm.Clang` VS component to be added to the VS 2019 image?

kaylangan:
@reupen thanks for pointing that out! I've just opened #1297 to add it to the VS2019 image. We'll keep this issue open to track the other images.

PR #1298, opened 2019-10-03
Add Clang VS component group to VS 2019 image

reupen:
Relates to #270

Based on previous PRs, I assume this is all that is required, but please let me know if anything else is.

The component ID can be verified at https://docs.microsoft.com/en-gb/visualstudio/install/workload-component-id-vs-enterprise?view=vs-2019#desktop-development-with-c.

(Additionally, I noticed when running `Install-VS2019.ps1` locally that there seem to be some invalid component IDs in the script e.g. `Microsoft.VisualStudio.Component.VC.v141`.)


reupen:
This is a dupe of #1297 – the other one is a few minutes older so will close this one 🙂

Issue #1235, opened 2019-09-13
Run-Antivirus fails on WS2019-VS2019


samrobillard:
The packer build fails at the 'Run-Antivirus' step with the following error:

==> managed-image: Provisioning with powershell script: /home/sam/repo/azure-pipelines-image-generation/images/win/scripts/Installers/Vs2019/Run-Antivirus.ps1
    managed-image: Run antivirus
    managed-image: Scan starting...
    managed-image: CmdTool: Failed with hr = 0x800106BA. Check C:\Users\packer\AppData\Local\Temp\MpCmdRun.log for more information
    managed-image: Set antivirus parmeters                                                                                                                                                                                                                                      ==> managed-image: Set-MpPreference : Operation failed with the following error: 0x800106b5. Operation: Set-MpPreference. Target:                                                                                                                                               ==> managed-image: Scan_AvgCPULoadFactor.                                                                                                                                                                                                                                       ==> managed-image: At C:\Windows\Temp\script-5d7b939f-2e7e-7e8c-fd16-160525bff985.ps1:19 char:1                                                                                                                                                                                 ==> managed-image: + Set-MpPreference -ScanAvgCPULoadFactor 5 `
==> managed-image: + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                 ==> managed-image:     + CategoryInfo          : NotSpecified: (MSFT_MpPreference:root\Microsoft\...FT_MpPreference) [Set-MpPreference],                                                                                                                                        ==> managed-image:    CimException                                                                                                                                                                                                                                              ==> managed-image:     + FullyQualifiedErrorId : HRESULT 0x800106b5,Set-MpPreference                                                                                                                                                                                            ==> managed-image:                                                                                                                                                                                                                                                              ==> managed-image: Set-MpPreference : Operation failed with the following error: 0x800106b5. Operation: MpPreference. Target:                                                                                                                                                   ==> managed-image: ConfigListExtension.
[...]

PR #1300, opened 2019-10-07
Update-MpSignature before pop-location

yl2558:


yl2558:
This is a fix for the #1235 Run-Antivirus fails on VS2019. 
#159 also reported the same issue as above. There is a solution about adding Update-MpSignature will make vs2019 generates reliably.
The Update-MpSignature cmdlet updates the antimalware definitions with the latest definitions available on an update server. 
I tested the changes in this release and it succeeded. https://dev.azure.com/mseng/AzureDevOps/_releaseProgress?_a=release-pipeline-progress&releaseId=12564579

Issue #1307, opened 2019-10-10
Documentation for VS-2019 is incomplete re: supported .NET Core versions


laynew:
**Describe the bug**
I believe the documentation or the VS-2019 image is incomplete. According to the documentation the latest supported version of .NET Core is 2.2.105. However I was able to build a project with 3.0.100 using this image.

**Expected behavior**
The documentation for the image should indicate that .NET Core 3.x.x is supported

**Actual behavior**
The documentation indicates that  2.2.105 is the latest supported version

**Image impacted**
`images/win/Vs2019-Server2019-Readme.md`

alepauly:
Verified that the docs are missing info. Also affects Ubuntu images

hiyadav:
Thanks for pointing that out. @alepauly can you let me know if the doc files need to be updated manually or should they be updated using generated doc files from the production build ?

alepauly:
> 
> 
> Thanks for pointing that out. @alepauly can you let me know if the doc files need to be updated manually or should they be updated using generated doc files from the production build ?

@hiyadav The docs should be autogenerated and not updated manually. I think the scripts that install dotnet core and generate the output have some bugs since they are missing the latest. We should fix the scripts.

simonauner:
Same issue as https://github.com/microsoft/azure-pipelines-image-generation/issues/1301 (if you want to close either as duplicate)

hiyadav:
We are working on it. PR should be merged soon.

hiyadav:
We have updated the installer scripts for both Windows and Linux images, and should start reflecting latest SDK's (in image and documentation) from next release.

Issue #1301, opened 2019-10-08
dotnet CLI is 3.0 even though docs says 2.2


simonauner:
**Describe the bug**
We're running a build with `ubuntu-latest` that uses `dotnet` CLI, and up until recently that default version has been `2.2.x`. We noticed that sometime during last week this was bumped to `3.0.100`. This breaks our build because the behavior of `dotnet publish --output my-folder` is no longer relative to the project folder, but ends up in the root of the build. Fixable, yes, but it's a bit unexpected to bump the software without notice.

I guess the actual **bug** is that the documentation isn't correct anymore: 
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent => https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md

**Expected behavior**
I expect the default version of a software (in this case `dotnet` CLI) to be documented in the README above.

**Actual behavior**
`dotnet` CLI is running version `3.0.100` which is undocumented.

![image](https://user-images.githubusercontent.com/1343979/66367512-ccaf8f80-e994-11e9-9507-d40ce563d696.png)

Steps to reproduce:
1. Create a build with YAML file
2. Choose pool `ubuntu-latest`
3. Run a script task: `dotnet --version`

**Image impacted**
 `ubuntu-latest`

simonauner:
This got a bit worse yesterday, when also `windows-latest'` runs .NET Core 3.0. Now the `UseDotNet` task fails to upgrade to 3.x if 3.x is already the default SDK.

**How to replicate:**
1. Run a job with `vmImage: 'windows-latest`
2. Run `UseDotNet` task:
    ```yaml
    - task: UseDotNet@2
      displayName: 'Update .NET Core sdk'
      inputs:
        packageType: sdk
        version: 3.x
    ```

![use-dot-net](https://user-images.githubusercontent.com/1343979/66981038-ef3b5a00-f0b2-11e9-9bd6-147d08486841.png)

hiyadav:
We have updated the installer scripts for both Windows and Linux images, and should start reflecting latest SDK's (in image and documentation) from next release.

hiyadav:
@simonauner the UseDotNet task is failing for another reason, we have a PR to fix this #11577 
This PR shall fix the task for downloading latest 3.0 version.

PR #1315, opened 2019-10-15
Extracted dotnetcore-sdk.sh for common use in ubuntu1604 and 1804.

hiyadav:
#1307 #1301


images/linux/scripts/installers/1604/dotnetcore-sdk.sh
```diff
@@ -1,91 +0,0 @@
...
-    dotnet build
-    set +e
-    cd .. || exit
-    rm -rf "$sdk"
-}
-
-set -e
-
-echo "Determing if .NET Core ($LATEST_DOTNET_PACKAGE) is installed"
-if ! IsInstalled $LATEST_DOTNET_PACKAGE; then
```
bishal-pdMSFT: you might have to use this logic for installing latest version. in the 1804 script, it uses logic to install for 18.04 ubuntu. Check if that logic works also on 16.04 ubuntu.

hiyadav: Started a CI run for this branch to verify if the change is required. At first look it looked liked to be same for both as we add .Net Core registry to linux registry manager from where it should get the package.

Here is the CI run for Ubuntu16 using which we can determine this:
https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=10658919

bishal-pdMSFT: @hiyadav if you have ubunutu 16 VM or WSL, you can check by running these commands manually as well

images/linux/scripts/installers/dotnetcore-sdk.sh
```diff
@@ -34,6 +34,7 @@ set -e
 echo "Determing if .NET Core ($LATEST_DOTNET_PACKAGE) is installed"
 if ! IsInstalled $LATEST_DOTNET_PACKAGE; then
     echo "Could not find .NET Core ($LATEST_DOTNET_PACKAGE), installing..."
+    add-apt-repository universe
     curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
     mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg
     sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod bionic main" > /etc/apt/sources.list.d/dotnetdev.list'
```
bishal-pdMSFT: do you need to say ubuntu 16.04 name instead of bionic for getting package list?

images/linux/scripts/installers/1604/dotnetcore-sdk.sh
hiyadav: Thanks, doing it on WSL helped. updated accordingly

images/linux/scripts/installers/dotnetcore-sdk.sh
hiyadav: did that thanks !

images/linux/ubuntu1604.json
```diff
@@ -126,7 +126,7 @@
                 "{{template_dir}}/scripts/installers/docker-compose.sh",
                 "{{template_dir}}/scripts/installers/docker-moby.sh",
                 "{{template_dir}}/scripts/installers/docker.sh",
-                "{{template_dir}}/scripts/installers/1604/dotnetcore-sdk.sh",
+                "{{template_dir}}/scripts/installers/dotnetcore-sdk1604.sh",
```
bishal-pdMSFT: why not go back to keeping these files in their own folders?

PR #1336, opened 2019-10-22
update readme for ubuntu16 20190930

yl2558:


DanielHabenicht:
Hi guys, are you planning on showing this Readme on official channels and in azure pipelines? Currently, it is pure speculation which software is running on the agent you are building on. Also, the time frame of commits, releases and announcements does not quite match up: https://docs.microsoft.com/en-us/azure/devops/release-notes/2019/sprint-159-update#updates-to-hosted-pipelines-images

PR #1354, opened 2019-10-28
Remove sbt validation since it's blocking prod generation.

alepauly:
The call to sbt --script-version in the Validate-Sbt.ps1 script is leading to a Robocopy call that's taking many hours and eventually timing out on error. Remove for now and to help investigate to unblock production deployments.

### DIFF
--- a/images/win/vs2017-Server2016-Azure.json
+++ b/images/win/vs2017-Server2016-Azure.json
@@ -619,12 +619,6 @@
         {
             "type": "powershell",
             "scripts":[
-                "{{ template_dir }}/scripts/Installers/Validate-Sbt.ps1"
-            ]
-        },
-        {
-            "type": "powershell",
-            "scripts":[
                 "{{ template_dir }}/scripts/Installers/Validate-Svn.ps1"
             ]
         },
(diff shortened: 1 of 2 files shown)


shadyysf:
@alepauly , so what exactly is the plan here? Just avoid the failing validation step for now? How are we going to investigate this? Also, is there something tracking the investigation work and making sure we re-enable after the investigation?

alepauly:
> 
> 
> @alepauly , so what exactly is the plan here? Just avoid the failing validation step for now? How are we going to investigate this? Also, is there something tracking the investigation work and making sure we re-enable after the investigation?

my plan is to reach out to the contributor that added these so that he can use the state of the machine to generate good validation. Is not ideal, and it's making me think I should remove the Install- step too...right? @shadyysf

shadyysf:
@alepauly , I'm fine either way. As long as we have something tracking this work so we don't forget about it and leave the tool either uninstalled or unverified.

alepauly:
> 
> 
> @alepauly , I'm fine either way. As long as we have something tracking this work so we don't forget about it and leave the tool either uninstalled or unverified.

I think it makes sense to remove the whole thing and reach out to the contributor to recheck.

Issue #496, opened 2018-11-21
Languages: Scala sbt and sbt-extras (Windows)


davidstaheli:
Languages: Scala sbt and sbt-extras (Windows)

tomvo12:
We also need sbt capable build agents to support our big data projects. Any idea, when this would be available?

DavidStaheli:
Hi @tomvo12.  With the upcoming holidays it might not be until January.  Will that disrupt your plans?  Can you get by with installing it at the beginning of your builds until then?
```
brew install sbt
```
Or:
```
choco install sbt
```

tomvo12:
Hi @DavidStaheli .
Thanks for the reply. If sbt would be available in the January 2019 release all is fine for us.

dkmiller:
Currently, only way I know of to run SBT in the default agents is by installing it on the fly, either via brew or chocolatey on a "physical" agent, or via downloading a Docker container containing SBT. Having a simple build task which wraps SBT would make things _much_ easier.

fergalsweeney:
We are looking to move to Azure Devops to build and deploy Scala app to AWS..I wish the hosted build servers supported SBT natively

BlackTurtle123:
I also want to buil Scala apps on Azure

quintindk:
I too would like a SBT build option in the hosted agents... Maybe someone has a how too with Docker as mentioned by @dkmiller

> Currently, only way I know of to run SBT in the default agents is by installing it on the fly, either via brew or chocolatey on a "physical" agent, or via downloading a Docker container containing SBT. Having a simple build task which wraps SBT would make things _much_ easier.

er1c:
🙏

er1c:
I took a stab at this via the PR: https://github.com/microsoft/azure-pipelines-image-generation/pull/1255 - happy to have other eyeballs on it :)

jacwalte:
@er1c Looks like your PR was merged - any documentation on use?

er1c:
@jacwalte honestly, I have no idea lol, this was my "first step" into the project.  Need to figure out how the old sbt was used.


@microsoft this should be resolved as "fixed"

er1c:
@jacwalte  these worked for me:

```yaml
# Starter pipeline
# Start with a minimal pipeline that you can customize to build and deploy your code.
# Add steps that build, run tests, deploy, and more:
# https://aka.ms/yaml
trigger:
- master
pool:
[...]

PR #1361, opened 2019-10-30
Add sbt scripts back to Windows packer scripts

alepauly:
The validation of sbt was causing build failures so I've added back the steps to install it but remove the specific lines during validation that were causing issues.


er1c:
@alepauly  if you are inclined, you can mark https://github.com/microsoft/azure-pipelines-image-generation/issues/496 as resolved/fixed/etc

PR #1367, opened 2019-11-01
add commit url to commit.txt file

yl2558:
Fix this [issue](https://github.com/github/pe-actions-compute/issues/149).

### DIFF
--- a/images/linux/ubuntu1604.json
+++ b/images/linux/ubuntu1604.json
@@ -1,6 +1,5 @@
 {
     "variables": {
-        "commit_id": "{{env `COMMIT_ID`}}",
         "commit_url": "{{env `COMMIT_URL`}}",
         "client_id": "{{env `ARM_CLIENT_ID`}}",
         "client_secret": "{{env `ARM_CLIENT_SECRET`}}",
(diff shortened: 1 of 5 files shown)


images/linux/ubuntu1604.json
```diff
@@ -55,6 +56,7 @@
                 "mkdir {{user `image_folder`}}",
                 "chmod 777 {{user `image_folder`}}",
                 "echo {{user `commit_id`}} > {{user `commit_file`}}",
+                "echo {{user `commit_url`}} > {{user `commit_file`}}",
```
alepauly: pretty sure the use of > will overwrite the contents of the previous line. do you need the previous line? might as well just remove it.

images/win/WindowsContainer1803-Azure.json
```diff
@@ -21,6 +21,7 @@
         "metadata_file": "C:\\image\\metadata.txt",
         "helper_script_folder": "C:\\Program Files\\WindowsPowerShell\\Modules\\",
         "commit_id": "LATEST",
+        "commit_url": "LATEST",
```
alepauly: since this is supposed to be a URL, I think you can just leave it empty.

images/linux/ubuntu1604.json
alepauly: actually, instead of adding a new parameter, what about just building the url in here using the provided commit_id? should be pretty simple. so modify the line before to not just output the commitid but the url built off the commit_id.

yl2558: The commit_id has been removed. Pass in the commit_url instead.

Issue #1015, opened 2019-06-11
Service Fabric needs to be upgraded again


jackbond:
**Describe the bug**
Can't build service fabric apps which target latest SF SDK.

**Expected behavior**
Can build service fabric apps which target latest SF SDK.

**Actual behavior**
Can't build service fabric apps which target latest SF SDK.

1.) Attempt to build service fabric app which targets latest SF SDK

**Image impacted**
Hosted Windows 2019 with VS2019 image

jstevans:
Ping -- @sachinma / @bishal-pdMSFT, are you still the right contacts to perform this upgrade, per the comment by @RoopeshNair in #213?

We need 6.5 CU 2 or later (preferably the latest, [which is currently 6.5CU5][sf-6-5-cu5]) in order to [make use of MITS][mits].

[sf-6-5-cu5];https://github.com/Azure/service-fabric/blob/master/release_notes/Service_Fabric_ReleaseNotes_65CU5.md
[mits]:https://docs.microsoft.com/en-us/azure/service-fabric/how-to-deploy-service-fabric-application-system-assigned-managed-identity

vinodkumar3:
We've figured out the issue, we're now working with Azure Pipeline: Hosted Agents team for the fix.
Thanks for raising this issue.

vinodkumar3:
Issue has been fixed and will be available in upcoming release.

PR #1389, opened 2019-11-08
Overriding main WebProductList.xml feed of WebPI

vinodkumar3:
In favor of Issue - #1015 
Currently during image generation ([here](https://github.com/microsoft/azure-pipelines-image-generation/blob/02cdd72e2bd058ae3d078d0b5d71d0240c994f2c/images/win/scripts/Installers/Vs2017/Initialize-VM.ps1#L102)) we are getting WebpiCmd [package ](https://chocolatey.org/packages/webpicmd)from chocolatey which is outdated hence corresponding feed is also outdated due to which newly generated images doesn't contain newly released Service Fabric SDKs.
In the following PR, I'm overriding the default feed to latest feed.


alepauly:
CI verified.

Issue #667, opened 2019-01-29
Add well-known SSH host keys to known_hosts


congyiwu:
**Describe the bug**
Unless a pipeline explicitly modifies known_hosts or disables host key checking, accessing repos via SSH will fail.  One customer reported that it hangs indefinitely while prompting for verification.

https://man.openbsd.org/ssh.1 mentions that both ~/.ssh/known_hosts and /etc/ssh/ssh_known_hosts are read.  For Mac and Linux images, it looks like we could edit known_hosts at image generation time.

On Windows, The Azure Pipelines Agent provides a copy of MinGit, which provides a copy of SSH.  I ran procmon and it checks:
* C:\Users\congyiw\.ssh\known_hosts
* C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\etc\ssh\ssh_known_hosts (which is under the MinGit folder)

So on Windows, I'm not sure if we can fix this in the image itself.

**Expected behavior**
Accessing repos via SSH on Azure Repos from Azure Pipelines should succeed :).  I would expect the same for other Git providers that we integrate tightly with, like GitHub and BitBucket.  That is, unless users generally expect 100% control over their known_hosts file.

**Actual behavior**
1. Add a shell script to your pipeline that does `git ls-remote <ssh remote url>`.
2. Queue the pipeline.

**Image impacted**
All

chrisrpatterson:
Given we don't really have support for SSH as part of our Source integration I am not sure we should do anything here. If we explicitly add support for SSH or recommend SSH for certain scenarios like submodules we would need to take this into account.  

A user can easily run ssh-keyscan to update their known-hosts file.

congyiwu:
> Given we don't really have support for SSH as part of our Source integration...

Users are already cloning with SSH from pipelines anyway, for submodules and also from custom steps for things like CocoaPods and Terraform, regardless of whether we explicitly recommend it.  To be fair, the percentage of users doing this could be trivial [1], so I'm fine with deferring this issue until there's obvious demand.

> A user can easily run ssh-keyscan to update their known-hosts file.

However, without explicitly verifying the contents of known_files afterwards, there's no guarantee the contents are correct.  For hosted agents scoped to individual builds [2], ssh-keyscan will run from scratch for each build.  If there is a MITM attack, the agents will blindly accept the attacker's host key.

[...]

PR #1395, opened 2019-11-12
Add known hosts

lupine86:
Unless a pipeline explicitly modifies known_hosts or disables host key checking, accessing repos via SSH will fail. 

This change sets the public keys for the GitHub and Azure DevOps at image generation time for Linux images.

See #667 for more details.


images/linux/scripts/installers/git.sh
```diff
@@ -30,6 +30,22 @@ if ! command -v git-lfs; then
     exit 1
 fi
 
+# Add well-known SSH host keys to known_hosts
+
+# Azure Devops fingerprint SHA256:ohD8VZEXGWo6Ez8GSEJQ9WpafgLFsOfLOtGGQCQo6Og
+echo ssh.dev.azure.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOfGJ4NakVyIzf1rXYd4d7wo6jBlkLvCA4odBlL0mDUyZ0/QUfTTqeu+tm22gOsv+VrVTMk6vwRU75gY/y9ut5Mb3bR5BV58dKXyq9A9UeB5Cakehn5Zgm6x1mKoVyf+FFn26iYqXJRgzIZZcZ5V6hrE0Qg39kZm4az48o0AUbf6Sp4SLdvnuMa2sVNwHBboS7EJkm57XQPVU3/QpyNLHbWDdzwtrlS+ez30S3AdYhLKEOxAG8weOnyrtLJAUen9mTkol8oII1edf7mWWbWVf0nBmly21+nZcmCTISQBtdcyPaEno7fFQMDD26/s0lfKob4Kw8H >> azureDevOpsPublicKey
+ssh-keygen -H -f azureDevOpsPublicKey
```
congyiwu: `-H` hashes the host name, right?  Why do we need to do that?
* The hostname and the fact that we want to support these hosts, is should be public information
* As a user, if I see a hashed hostname in a known_hosts file that someone else gave me, how would I know to trust it?

lupine86: [HashKnownHosts ](https://linux.die.net/man/5/ssh_config)is enabled on Ubuntu by default in the ssh_config.  If we don't hash the public key hostname ssh won't read it correctly as a known host and will just add it again if I'm not mistaken.

lupine86: Never mind.  I just tested and ssh seems to work fine when the hostnames are not hashed even when HashKnownHosts is enabled.  I will remove this.

congyiwu: Ah, I didn't realize Ubuntu changed the default.  I was curious about why, and it turns out it's inherited from Debian It's intended to prevent worms from spreading to targets in the known_hosts file:
https://marc.info/?l=openssh-unix-dev&m=111645649404089&w=2
https://lists.debian.org/debian-devel/2005/07/msg00041.html

I don't think the arguments apply to the entries that we're adding, so it's fine to not hash since it still works.

If you do need to hash, you should be able to run `ssh-keygen -H -f /etc/ssh/ssh_known_hosts` once directly, instead of on intermediate files.

images/linux/scripts/installers/git.sh
```diff
@@ -30,6 +30,18 @@ if ! command -v git-lfs; then
     exit 1
 fi
 
+# Add well-known SSH host keys to known_hosts
+
+# Azure Devops fingerprint SHA256:ohD8VZEXGWo6Ez8GSEJQ9WpafgLFsOfLOtGGQCQo6Og
+echo ssh.dev.azure.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC7Hr1oTWqNqOlzGJOfGJ4NakVyIzf1rXYd4d7wo6jBlkLvCA4odBlL0mDUyZ0/QUfTTqeu+tm22gOsv+VrVTMk6vwRU75gY/y9ut5Mb3bR5BV58dKXyq9A9UeB5Cakehn5Zgm6x1mKoVyf+FFn26iYqXJRgzIZZcZ5V6hrE0Qg39kZm4az48o0AUbf6Sp4SLdvnuMa2sVNwHBboS7EJkm57XQPVU3/QpyNLHbWDdzwtrlS+ez30S3AdYhLKEOxAG8weOnyrtLJAUen9mTkol8oII1edf7mWWbWVf0nBmly21+nZcmCTISQBtdcyPaEno7fFQMDD26/s0lfKob4Kw8H >> azureDevOpsPublicKey
```
congyiwu: Can you `echo` directly into `/etc/ssh/ssh_known_hosts`?

lupine86:
Abandoning these changes.  The changes were added to the Provisioner factory scripts instead.  See https://dev.azure.com/mseng/AzureDevOps/_git/MMS.Provisioner/pullrequest/520944.

PR #1418, opened 2019-11-26
[WIP] Install PyPy and Boost toolcache via NPM

ivan-nosar:
This PR replaces installation of PyPy in tool cache with NPM approach. 
Affected images:
- Ubuntu-1604
- Ubuntu-1804
- VS-2017
- VS-2019


AlenaSviridenko:
Not sure if it is possible but maybe we could find or create a common place to store PyPy available versions? because now we have 3 files specifying PyPy versions to be installed and every new version will force engineers to add them to 3 different files.
Please, let me know if such implementation potentially can bring a lot of breaking changes and it is not worth it.

maxim-lobanov:
> Not sure if it is possible but maybe we could find or create a common place to store PyPy available versions? because now we have 3 files specifying PyPy versions to be installed and every new version will force engineers to add them to 3 different files.
> Please, let me know if such implementation potentially can bring a lot of breaking changes and it is not worth it.

Agree, let's place the list of versions to some common place because it is the same through all images.

Also I would recommend specifying only major version for PyPy because we have only two versions now:
```
    npm install toolcache-pypy-ubuntu-1804-x64@2
    npm install toolcache-pypy-ubuntu-1804-x64@3
```

images/win/toolcache.json
```diff
@@ -0,0 +1,6 @@
+{
+    "toolcache-pypy-windows-x86": [
+        "2", "3"
+    ],
+    "toolcache-boost-windows-x64": [ "1.69" ]
+}
```
AlenaSviridenko: this way you still have 2 separate files with tools versions, although versions numbers are the same. 
Maybe we could place versions to some root folder like `azure-pipelines-image-generation/images/` but leave specific packages names specified in platform files, like you did for Boost on Ubuntu https://github.com/microsoft/azure-pipelines-image-generation/pull/1418/files#diff-ac45f01c6fccdf6b0b01ab644c8a4b22R50 ?

ivan-nosar: The version sets may be different through all the images (e.g. we may need to install Boost 1.70 on Windows, but none on Ubuntu). Moreover I really don't think that maintaining of few `.json` files stored in different places is so good idea...

AlenaSviridenko: > Moreover I really don't think that maintaining of few .json files stored in different places is so good idea...

I agree, that is why I was asking about single `.json` file in the root.
Anyway, if it is real case to have different sets of versions for different OS types - then I agree with different files for Windows and Ubuntu.

images/linux/scripts/installers/1604/hosted-tool-cache.sh
```diff
@@ -28,12 +32,36 @@ for setup in $setups; do
...
+chmod -R 777 $AGENT_TOOLSDIRECTORY
+
+echo "Installing npm-toolcache..."
+
+toolVersionsFileContent=$(cat "$INSTALLER_SCRIPT_FOLDER/toolcache.json")
+
+tools=$(echo $toolVersionsFileContent | jq -r 'keys | .[]')
+
+for tool in ${tools[@]}; do
+    toolVersions=$(echo $toolVersionsFileContent | jq -r ".[\"$tool\"] | .[]")
```
maxim-lobanov: You can just pass file name to jq so you don't need `toolVersionsFileContent=$(cat "$INSTALLER_SCRIPT_FOLDER/toolcache.json")`

`jq -r ".[\"$tool\"] | .[]" "$INSTALLER_SCRIPT_FOLDER/toolcache.json"`

images/win/scripts/Installers/Download-ToolCache.ps1
```diff
@@ -4,49 +4,81 @@
...
 }
 
-Set-Location -Path $current
+Pop-Location
 
+# Define AGENT_TOOLSDIRECTORY environment variable
+$env:AGENT_TOOLSDIRECTORY = $ToolsDirectory
 setx AGENT_TOOLSDIRECTORY $ToolsDirectory /M
 
+# Install tools form NPM
```
alepauly: nit: from

images/linux/toolcache.json
```diff
@@ -0,0 +1,6 @@
+{
+    "toolcache-pypy-ubuntu-%s-x64@%s": [
```
alepauly: I think we should start getting rid of the toolcache name. This are just common tools that are preinstalled IMO. The file should probably be toolpackages.json? and the package names can just skip the "toolcache" part. @shadyysf thoughts?

shadyysf: @alepauly , makes sense to me. Let's drop the "toolcache" part from names,paths... etc.

PR #1424, opened 2019-11-30
Update cmake.sh to 3.15 stable

merlinLia:
`cmake 3.15.5 `has been released in 2019.10.30. Update CMake to 3.15 stable could let us to make it possible to apply the latest CMake in our  build.

### DIFF
--- a/images/linux/scripts/installers/cmake.sh
+++ b/images/linux/scripts/installers/cmake.sh
@@ -13,7 +13,7 @@ echo "Checking to see if the installer script has already been run"
 if command -v cmake; then
     echo "Example variable already set to $EXAMPLE_VAR"
 else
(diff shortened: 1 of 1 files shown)


crazyhappygame:
Latest cmake release is 3.16 https://cmake.org/download/

LiangMengdi:
@crazyhappygame Afraid its [stability ](https://gitlab.kitware.com/cmake/cmake/issues) since the `cmake 3.16.0` was just published not long ago. So, here want the `cmake` update to the latest stable version. Will update to 3.16.* some days later.

kaylangan:
@merlinLia thanks for the contribution! Could you open a PR in https://github.com/actions/virtual-environments? We've moved the development of the Ubuntu images there.

PR #1428, opened 2019-12-03
add VS2019 component ArchitectureToolsManaged

coolbry95:
This was added in VS2017. This is causing an upgrade to VS2019 pipelines to fail due to not being able to build without the missing component.


kaylangan:
Thanks for the contribution! Could you open a PR in https://github.com/actions/virtual-environments? We've moved development of the Windows images there.

PR #1465, opened 2020-01-16
Install Windows Updates during WinCon build.

alepauly:
We need to make sure the image is fully patched since there can be replication delays from base image in Azure Gallery.


alepauly:
Low risk force merging to include in new build.

PR #1475, opened 2020-02-03
Fixing chocolatey installation

al-cheb:
Issue:
Unable to install chocolatey - The request was aborted: Could not create SSL/TLS secure channel.

Fix:
Enable Tls1.2.

### DIFF
--- a/images/win/scripts/Installers/WindowsContainer1803/Initialize-VM.ps1
+++ b/images/win/scripts/Installers/WindowsContainer1803/Initialize-VM.ps1
@@ -72,6 +72,7 @@ else {
 }
 
 # Run the installer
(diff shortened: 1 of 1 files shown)


ferventcoder:
Just a note, we are updating the documentation installation pages as a result of the change we made this morning. I'm not sure you will run into this but we went by integer instead of enum values for the use case where someone has loaded PowerShell v3 which uses .NET 4.x, but has .NET 4.5+ installed.

`[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072` It won't know of the enumeration value, but the integer set will work.

PR #1493, opened 2020-03-04
Update README.md

WilBrady:
Modify readme to be clear the use of this repository.


WilBrady:
@sergey-akhalkov - I love that idea! we should definitely have a bot close issues with a comment that active image development is happening on virtual environments (something like that). If you don't mind making that happen, please do!

PR #1516, opened 2020-04-09
Clarify that docker is the client, not the server

EugenMayer:
Also see discussion under https://developercommunity.visualstudio.com/content/problem/984112/docker-under-macos-1014.html


EugenMayer:
Well i guess someone just ignore this PR and did his own on top :) Collaboration at it's best, do work twice and throw away one :)

Closing this one then

PR #1525, opened 2020-06-06
Remove seemingly unused file `q`

sorpaas:
I don't see the purpose of `q` and why it should be in this repository.


al-cheb:
@sorpaas, Hello.

We have stopped maintain  `azure-pipelines-image-generation` and migrated the project to the new repository - https://github.com/actions/virtual-environments. In that case, we should not update or delete any files. If you would like to contribute, please use the https://github.com/actions/virtual-environments repo.

PR #1528, opened 2020-09-13
Update Install-NodeLts.ps1

asklar:
Related: https://github.com/actions/virtual-environments/pull/1566

yarn is able to create a duplicate environment variable by using different casing in the env block it passes to CreateProcess. This creates problems for downstream tools that expect exactly one version. Until this is fixed in yarn (https://github.com/yarnpkg/yarn/pull/8337) we should only set the lowercase variable name


asklar:
CC @maxim-lobanov

